PR #16, opened 2018-06-13
fix:returns null when block not found instead returns 405

dinn2018:


qianbin:
it's a panic recovered by http server

PR #62, opened 2018-07-03
feat(API): timeout by adding accounts call context

dinn2018:
accounts call context

### DIFF
--- a/api/accounts/accounts.go
+++ b/api/accounts/accounts.go
@@ -6,9 +6,11 @@
 package accounts
 
 import (
+	"context"
 	"math/big"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/math"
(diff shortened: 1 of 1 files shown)


api/accounts/accounts.go
```diff
@@ -142,19 +143,32 @@ func (a *Accounts) Call(to *thor.Address, body *ContractCall, header *block.Head
...
-		return nil, err
-	}
-	if err := state.Err(); err != nil {
-		return nil, err
+	vmout := make(chan *runtime.Output)
+	go func() {
+		o, _ := exec()
+		vmout <- o
+	}()
+	for {
```
qianbin: no for loop needed

api/accounts/accounts.go
```diff
@@ -142,19 +143,32 @@ func (a *Accounts) Call(to *thor.Address, body *ContractCall, header *block.Head
...
+		select {
+		case <-ctx.Done():
+			if err := ctx.Err(); err != nil {
```
qianbin: no check err, just return ctx.Err()

api/accounts/accounts.go
```diff
@@ -142,19 +143,32 @@ func (a *Accounts) Call(to *thor.Address, body *ContractCall, header *block.Head
 			GasLimit:    header.GasLimit(),
 			TotalScore:  header.TotalScore()})
 
-	vmout := rt.ExecuteClause(clause, 0, body.Gas, &xenv.TransactionContext{
+	exec, interrupted := rt.PrepareClause(clause, 0, body.Gas, &xenv.TransactionContext{
```
qianbin: rename 'interrupted' to 'interrupt'

PR #66, opened 2018-07-04
Create Dockerfile

zubietaroberto:
Inspired by #61 , which was not working on my environment. 
Added a Dockerfile based on Alpine Linux. Totals about 45MB. I am still struggling to make the exposed ports work.


libotony:
see #67

PR #71, opened 2018-07-09
Comment added to structure

hariexcel:
Added Note to structure


qianbin:
Thanks for your PR. That will be fine if more comments added.

qianbin:
incomplete commits

PR #127, opened 2018-08-27
Feat: chain BlockReader

uldaman:


chain/chain.go
```diff
@@ -539,3 +539,70 @@ func (c *Chain) IsBlockExist(err error) bool {
...
+			}
+
+			blocks = append(blocks, &Block{positionBlock, true})
+			position = positionBlock.Header().ParentID()
+		}
+	})
+}
+
+func (c *Chain) nextBlock(descendantID thor.Bytes32, num uint32) (*block.Block, error) {
+	next, err := c.GetAncestorBlockID(descendantID, num+1)
```
qianbin: dead lock

chain/chain.go
```diff
@@ -539,3 +539,70 @@ func (c *Chain) IsBlockExist(err error) bool {
...
+	if err != nil {
+		return nil, err
+	}
+
+	return c.GetBlock(next)
```
qianbin: dead lock

PR #251, opened 2019-05-16
Go mod

libotony:


libotony:
Added a new branch [go-mod](https://github.com/vechain/thor/tree/go-mod) for give it a try

PR #261, opened 2019-05-29
Fix transfer log

libotony:


qianbin:
I think no need to record wrong receipts root, because block id(hash) already includes it somehow.

PR #322, opened 2020-01-13
New storage backend and more...

qianbin:
Improvements include but not limited to:

* Deeply optimized state trie storage and cache, brings
    * grouped and ordered trie nodes store results in faster state trie access, at least 2X for OP_SLOAD;
    * dynamic state trie pruner saving 80% disk space occupation;
    * multi-level caches reduce 50% RAM usage under default setting, without impacting performance;
    * and much lower disk IO.

* Log db upgrade
    * newly designed table schemas compact db size to about 40% of previous.
    * optimized indexes cover most common cases to extremely speed up queries.

* Simplify block-chain store, by introducing the concept `Repository`, which is the container of chains.

* Improve state error handling, by explicitly returning errors on state access. Which makes the node more robust.


api/blocks/blocks.go
```diff
@@ -72,24 +71,29 @@ func (b *Blocks) parseRevision(revision string) (interface{}, error) {
...
 }
 
-func (b *Blocks) getBlock(revision interface{}) (*block.Block, error) {
+func (b *Blocks) getBlockSummary(revision interface{}) (s *chain.BlockSummary, err error) {
+	var id thor.Bytes32
 	switch revision.(type) {
 	case thor.Bytes32:
-		return b.chain.GetBlock(revision.(thor.Bytes32))
+		id = revision.(thor.Bytes32)
+		return b.repo.GetBlockSummary(revision.(thor.Bytes32))
```
libotony: better remove L79

qianbin: done with f5c7fea70b2ac7260eb3404a453a47147bbd94cb

PR #324, opened 2020-01-16
add more flags to gain more control in solo mode.

laalaguer:
Flags Added:
- skip logs flag.
- tx pool limit of length flag.
- tx pool limit of tx from the same account flag.


cmd/thor/flags.go
```diff
@@ -128,4 +128,14 @@ var (
 		Name:  "skip-txpool-blocklist",
 		Usage: "skip applying txpool blocklist",
 	}
+	txPoolLengthFlag = cli.IntFlag{
+		Name:  "txpool-length",
```
qianbin: suggested to be "txpool-limit"

cmd/thor/flags.go
```diff
@@ -128,4 +128,14 @@ var (
...
+		Value: 10000,
+		Usage: "set txpool length",
+	}
+	txPoolLimitAccountFlag = cli.IntFlag{
+		Name:  "txpool-limit-account",
```
qianbin: suggested to be "txpool-limit-per-account"

cmd/thor/flags.go
laalaguer: fixed.

cmd/thor/flags.go
laalaguer: fixed.

laalaguer:
`skipLogs` flag will be fixed in another pull request, hence removed from the current pull request.

PR #356, opened 2020-06-08
extend block: add bakcer signature, backer signature root,proposal

libotony:


block/block.go
```diff
@@ -57,34 +63,71 @@ func (b *Block) Transactions() tx.Transactions {
...
+		backerSignatures BackerSignatures
+	)
+
+	if err := s.Decode(&raws); err != nil {
 		return err
 	}
+	if len(raws) > 3 {
+		return errors.New("rlp:block body has too many fields")
+	}
+	if err := rlp.Decode(bytes.NewReader(raws[0]), &header); err != nil {
```
qianbin, line 113: potential out of bound attack

block/block.go
```diff
@@ -57,34 +63,71 @@ func (b *Block) Transactions() tx.Transactions {
...
-	return &Body{append(tx.Transactions(nil), b.txs...)}
+	return &Body{
+		append(tx.Transactions(nil), b.txs...),
+		append(BackerSignatures(nil), b.backerSignatures...),
+	}
 }
 
 // EncodeRLP implements rlp.Encoder.
 func (b *Block) EncodeRLP(w io.Writer) error {
+	if b.backerSignatures == nil {
```
qianbin, line 81: should use len() == 0 instead of comparing with nil

PR #362, opened 2020-07-06
consensus: validate block's backer signatures

libotony:


block/backer_signature.go
```diff
@@ -60,7 +60,7 @@ func (bs *BackerSignature) Validate(alpha []byte) (beta []byte, err error) {
 	defer func() { bs.cache.beta.Store(beta) }()
 
 	if len(bs.body) != 81+33 {
-		return []byte{}, errors.New("invalid backer signature length, 113 bytes needed")
+		return []byte{}, errors.New("invalid backer signature length, 114 bytes needed")
```
qianbin, line 63: `return nil, ...`

consensus/validator.go
```diff
@@ -196,6 +208,62 @@ func (c *Consensus) validateBlockBody(blk *block.Block) error {
 		}
 	}
 
+	bss := blk.BackerSignatures()
```
qianbin, line 211: can following lines be put inside vip193 check?

consensus/validator.go
```diff
@@ -196,6 +208,62 @@ func (c *Consensus) validateBlockBody(blk *block.Block) error {
...
+	if header.BackerSignaturesRoot() != bss.RootHash() {
+		return consensusError(fmt.Sprintf("block backers root mismatch: want %v, have %v", header.BackerSignaturesRoot(), bss.RootHash()))
+	}
+
+	totalBackers := uint64(len(bss)) + parent.TotalBackersCount()
+	if header.TotalBackersCount() != totalBackers {
+		return consensusError(fmt.Sprintf("block total backers count invalid: want %v, have %v", header.TotalBackersCount(), totalBackers))
+	}
+
```
qianbin, line 220: discussion: is there a window that one can build a vip193 block to split the network? if so, how to prevent it

consensus/validator.go
```diff
@@ -196,6 +208,62 @@ func (c *Consensus) validateBlockBody(blk *block.Block) error {
 		}
 	}
 
+	bss := blk.BackerSignatures()
+	if header.BackerSignaturesRoot() != bss.RootHash() {
+		return consensusError(fmt.Sprintf("block backers root mismatch: want %v, have %v", header.BackerSignaturesRoot(), bss.RootHash()))
+	}
+
+	totalBackers := uint64(len(bss)) + parent.TotalBackersCount()
```
qianbin, line 216: can be put in if statement.
less vars is better

consensus/validator.go
```diff
@@ -196,6 +208,62 @@ func (c *Consensus) validateBlockBody(blk *block.Block) error {
...
+			return consensusError("invalid block: backer signatures should be empty before fork VIP193")
+		}
+	} else {
+		leader, _ := header.Signer()
+		if len(bss) > 0 {
+			proposers, err := candidates.Pick(state)
+			if err != nil {
+				return err
+			}
+			all := make(map[thor.Address]bool, len(proposers))
```
qianbin, line 232: 可能不划算, 直接循环查找？

consensus/validator.go
```diff
@@ -196,6 +208,62 @@ func (c *Consensus) validateBlockBody(blk *block.Block) error {
...
+			for _, p := range proposers {
+				all[p.Address] = true
+			}
+
+			knownBackers := make(map[thor.Address]bool, len(bss))
```
qianbin, line 237: if bss are sorted, no need to make a set. just compare bs address with previous

PR #368, opened 2020-07-14
feature: local submitted tx exempt from wash out.

laalaguer:


txpool/tx_pool.go
```diff
@@ -400,25 +401,60 @@ func (p *TxPool) wash(headBlock *block.Header) (executables tx.Transactions, rem
...
+	// this limit only apply to non-locally submitted txs.
+	// locally submitted are exempt from this limit.
 	limit := p.options.Limit
 
-	// remove over limit txs, from non-executables to low priced
-	if len(executableObjs) > limit {
-		for _, txObj := range nonExecutableObjs {
-			toRemove = append(toRemove, txObj)
-			log.Debug("non-executable tx washed out due to pool limit", "id", txObj.ID())
+	// sort objs by local first, non-local later.
```
qianbin, line 408: to reduce complexity, it's recommended to follow the old procedure. don't distinguish  local with non-local in early stage, but later. pseudo-code like:

```
total := len(executableObjs) + len(nonExecutableObjs)
washedOutCount := 0
for i:= len(nonExecutableObjs)-1; i>=0 && total - washedOutCount > limit  ; i-- {
     obj:= nonExecutableObjs[I]
    if !obj.IsLocalSubmitted() {
        washedOutCount++
        toRemove = append(toRemove, obj)
   }
}

// same logic for executableObjs
...
```

then, we'll find that the return value `executables` is incorrect. it can be resolved by re-type `toRemove` to `map[*txObject]bool`, which helps filter out `executables` from `executableObjs`.

laalaguer: Or, we simply intervene earlier in the loop and only loop for once with O(n) complexity. See commit below.

qianbin:
another suggestion, how about adding a new method `AddLocal` instead of new argument to `Add`? If so, less code will be affected.

PR #406, opened 2020-10-27
node: add backer loop

libotony:


cmd/thor/node/backer_loop.go
```diff
@@ -0,0 +1,427 @@
...
+	var scope event.SubscriptionScope
+	defer scope.Close()
+
+	newProposalCh := make(chan *comm.NewProposalEvent)
+	scope.Track(n.comm.SubscribeProposal(newProposalCh))
+
+	newAcceptedCh := make(chan *comm.NewAcceptedEvent)
+	scope.Track(n.comm.SubscribeAccepted(newAcceptedCh))
+
+	unknownTicker := time.NewTicker(time.Duration(1) * time.Second)
```
qianbin, line 142: n * time.Second is ok

cmd/thor/node/backer_loop.go
```diff
@@ -0,0 +1,427 @@
...
+				}
+
+				if err := n.validateBackerSignature(ev.Signature, p, st); err != nil {
+					log.Debug("failed to validate backer signature", "err", err)
+					continue
+				}
+
+				n.comm.BroadcastAccepted(ev.Accepted)
+			} else {
+				unknownAccepted.Set(ev.Hash(), ev.Accepted)
```
qianbin, line 241: should be validated here?

cmd/thor/node/backer_loop.go
```diff
@@ -0,0 +1,427 @@
...
+				continue
+			}
+			seenProposal.Add(hash, struct{}{})
+
+			if parent.Header().ID() != p.ParentID {
+				unknownProposal.Set(hash, ev.Proposal)
+				continue
+			}
+
+			if st == nil || st.ParentID() != parent.Header().ID() {
```
qianbin, line 173: it's better to maintain bestblock(parent) and st under unknownTicker only.

cmd/thor/node/backer_loop.go
libotony: The backer signature can not be validated without the proposal

PR #409, opened 2020-10-30
node: clarify re-schedle condition

libotony:


zzGHzz:
2. best block is better than the block to be proposed

zzGHzz (approved):

PR #565, opened 2023-03-20
Add API call for public access to the pending transaction pool

picoAMPERE:
Currently there is no direct way to access the transaction pool through the API. Adding this functionality will enable users to query information about transactions in the tx pool. Currently the information is only available through modified nodes and providing a public access will make it available to everyone. It is a common function on other EVM based networks as well.

Changes:
1. Added functionality to `transaction.go` for handling the API request and retrieving all pending transactions from the tx pool.
  - New RESTful API endpoint with query parameters: 
    - /transactions/pending: Returns transaction IDs from the tx pool.
    - /transactions/pending?raw=true: Returns raw transactions from the tx pool.
    - /transactions/pending?expanded=true: Returns detailed transactions from the tx pool.
2. Implemented tests for pending transactions.
3. Added documentation for querying the tx pool with swagger

Note: `thor.yaml` has been extended but needs to be compiled as a binary version into `bindata.go` to make the updated documentation available.

### DIFF
--- a/api/transactions/transactions.go
+++ b/api/transactions/transactions.go
@@ -226,18 +226,18 @@ func (t *Transactions) getAllTxIDsFromTxpool() ([]string, error) {
 
 func (t *Transactions) handleGetAllTxIDsFromTxPool(w http.ResponseWriter, req *http.Request) error {
     raw := req.URL.Query().Get("raw")
-    detailed := req.URL.Query().Get("detailed")
+    expanded := req.URL.Query().Get("expanded")
 
     if raw != "" && raw != "false" && raw != "true" {
         return utils.BadRequest(errors.New("raw should be boolean"))
     }
 
-    if detailed != "" && detailed != "false" && detailed != "true" {
-        return utils.BadRequest(errors.New("detailed should be boolean"))
+    if expanded != "" && expanded != "false" && expanded != "true" {
+        return utils.BadRequest(errors.New("expanded should be boolean"))
     }
 
     rawBool := raw == "true"
-    detailedBool := detailed == "true"
+    expandedBool := expanded == "true"
 
     txs := t.pool.Executables()
     if rawBool {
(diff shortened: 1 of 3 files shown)


libotony:
Thank you for your contribution.It may be appropriate to consider relocating the given API endpoint to a more suitable path such as under path `/node` or `/node/txpool`, considering the variability in the status of the transaction pool across different nodes.

libotony:
> New RESTful API endpoint with query parameters:
>    /transactions/pending: Returns transaction IDs from the tx pool.
>    /transactions/pending?raw=true: Returns raw transactions from the tx pool.
>    /transactions/pending?expanded=true: Returns detailed transactions from the tx pool.

The usage of the parameter `raw` may not be required for optimal performance in managing the transaction pool. The value of the transaction pool is dynamic and can be subject to significant changes. The process of encoding transactions in this context can impose additional computing overheads. I would suggest removing `raw` from this API.

api/transactions/transactions.go
```diff
@@ -212,11 +212,64 @@ func (t *Transactions) parseHead(head string) (thor.Bytes32, error) {
 	}
 	return h, nil
 }
+ 
+func (t *Transactions) getAllTxIDsFromTxpool() ([]string, error) {
```
libotony, line 216: This function is not used.

api/transactions/transactions.go
```diff
@@ -212,11 +212,64 @@ func (t *Transactions) parseHead(head string) (thor.Bytes32, error) {
...
+    }
+
+    if expanded != "" && expanded != "false" && expanded != "true" {
+        return utils.BadRequest(errors.New("expanded should be boolean"))
+    }
+
+    rawBool := raw == "true"
+    expandedBool := expanded == "true"
+
+    txs := t.pool.Executables()
```
libotony, line 242: Use `t.pool.Dump()`. Correct me if I am wrong @qianbin.

libotony (changes requested):

api/transactions/transactions.go
qianbin: For pending transactions, `t.pool.Executables()` is fine.

picoAMPERE:
closing as txpool api was added in ~v2.1.0

PR #575, opened 2023-03-29
docker container runs thor under new user

kgapos:
docker runs the node though a new `thor` user, following [the slowmist security guidelines](https://github.com/slowmist/vechain-core-nodes-security-checklist/blob/master/README-en.md#23-avoid-using-root-to-run-thor), instead of the current implementation, which runs it under `root`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,8 +10,11 @@ RUN make all
 FROM alpine:latest
 
 RUN apk add --no-cache ca-certificates
-COPY --from=builder /go/thor/bin/thor /usr/local/bin/
-COPY --from=builder /go/thor/bin/disco /usr/local/bin/
+RUN adduser -D -s /bin/ash thor
+USER thor
+ENV PATH="$PATH:/home/thor"
+COPY --from=builder /go/thor/bin/thor /home/thor
+COPY --from=builder /go/thor/bin/disco /home/thor
 
 EXPOSE 8669 11235 11235/udp 55555/udp
 ENTRYPOINT ["thor"]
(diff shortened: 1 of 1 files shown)


kgapos:
This is a minor improvement, I suppose the blast radius is already limited by docker. I was more interested in trying out the contribution process. Feel free to merge this (or not) :shipit:

Dockerfile
```diff
@@ -10,8 +10,11 @@ RUN make all
 FROM alpine:latest
 
 RUN apk add --no-cache ca-certificates
-COPY --from=builder /go/thor/bin/thor /usr/local/bin/
-COPY --from=builder /go/thor/bin/disco /usr/local/bin/
+RUN adduser -D -s /bin/ash thor
+USER thor
+ENV PATH="$PATH:/home/thor"
+COPY --from=builder /go/thor/bin/thor /home/thor
```
libotony, line 16: Any particular concern about putting thor under the new user home directory other than `/usr/local/bin`?

kgapos: In a multi-user system, if a malicious user gains write access to `/usr/local/bin`, they could potentially replace system-wide executables with their own, leading to security issues. 

However, in a containerized environment, this concern is less relevant, as each container is isolated from others and runs a single application. Since this is not a multi-user environment, I will revert back to `/usr/local/bin` as suggested, thanks!

libotony:
Thanks！ Looks good to me, I'll test locally

libotony:
Wasted some on my local container runtime [colima](https://github.com/abiosoft/colima) which has some problems of mounted volume, Docker Desktop for mac and docker engine on linux works fine.

libotony (approved):

PR #588, opened 2023-05-04
doc: add hint of docker default user's change

kgapos:
Also remove broken link from ToC

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,6 @@ This is the first implementation written in golang.
 * [Installation](#installation)
   * [Requirements](#requirements)
   * [Getting the source](#getting-the-source)
-  * [Dependency management](#dependency-management)
   * [Building](#building)
 * [Running Thor](#running-thor)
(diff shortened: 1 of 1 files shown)


kgapos:
I came across this issue while upgrading my nodes. I reused the `/thor` volume I had created under `root` on the docker host. Naturally the new `thor` user in the container does not have permissions to that root directory by default. I guess this may throw off some users? Up to you if you want to merge this.

image

libotony:
In my understanding, `rw` should be fine.

libotony:
> > In my understanding, `rw` should be fine.
> 
> That was also my assumption, I tried `rw` and it didn't work. It works with `rwx` though. Perhaps the execute permission is required so that the thor user [can search the directory](http://web.mit.edu/sipb/doc/working/afs/html/subsection3.1.html)?

I just did a test, execute permission for folders means entering it thus you can read/write/traverse files under it.

PR #595, opened 2023-07-11
ignore known tx in pending tx subscription

libotony:


qianbin:
I think per-connection LRU is more reasonable, or let the client do the filtering.

libotony:
Client filtering might not be a good idea, since duplicated tx happens a lot, caused by multiple peer-to-peer node broadcasting new tx.

As a hub for re-distributing events, it might be efficient to filter in place. Per-connection LRU is perfect at the design level, but it might cost more memories, do you consider it's affordable?



> I think per-connection LRU is more reasonable, or let the client do the filtering.

qianbin:
> As a hub for re-distributing events, it might be efficient to filter in place. Per-connection LRU is perfect at the design level, but it might cost more memories, do you consider it affordable?

Sorry, I misunderstood the code.

A few minor changes are made within the last two commits.

libotony:
> > As a hub for re-distributing events, it might be efficient to filter in place. Per-connection LRU is perfect at the design level, but it might cost more memories, do you consider it affordable?
> 
> Sorry, I misunderstood the code.
> 
> A few minor changes are made within the last two commits.

Changes LGTM, still needs your review, I can't review it due to I am the author.

qianbin (approved):

PR #597, opened 2023-07-18
fix unexpected race in tx subscription

libotony:


qianbin:
Did not notice it's rwmutex. I think mutex is ok.

qianbin (approved):

PR #615, opened 2023-10-23
chore: unnecessary use of fmt.Sprintf

testwill:


libotony:
Hi @testwill, thank you for the contribution, package `discv5` is from our upstream go-etherem, as you can see from the [README](https://github.com/vechain/thor/tree/master/p2psrv/discv5#readme), as this is a minor issue, we would like to keep as is, thank you for your understanding.

PR #631, opened 2023-12-20
adding code owners

darrenvechain:
When this gets merged, we can add this setting:

`Require review from Code Owners`

image


claytonneal:
hiya, there is a team called "core-protocol" in github
could be nice to work this through the team, rather than personal github id's
thanks

image

libotony (approved):

libotony (approved):

PR #632, opened 2023-12-20
Tx pool tests

MakisChristou:
This PR fixes the [following issue](https://github.com/vechainfoundation/protocol-board-repo/issues/24) by increasing the coverage of all files to 80%. @libotony let me know if I should make any adjustments. Maybe you can recommend a better way to create a DevNet with a custom timestamp.


libotony:
@MakisChristou Could you do a rebase against the master branch? There is a major change recently of the import path.

txpool/tx_object_test.go
```diff
@@ -29,6 +29,13 @@ func newChainRepo(db *muxdb.MuxDB) *chain.Repository {
 	return repo
 }
 
+func newChainRepoCustomTimestamp(db *muxdb.MuxDB, timestamp uint64) *chain.Repository {
+	gene := genesis.NewDevnetCustomTimestamp(timestamp)
```
libotony, line 33: What's the purpose of new genesis with timestamp? Is it for the following code? https://github.com/vechain/thor/blob/352fbc6f4e1faf442e6efff255307b427093bc86/txpool/tx_pool.go#L238

MakisChristou:
> @MakisChristou Could you do a rebase against the master branch? There is a major change recently of the import path.

Hello @libotony . Regarding this, I am having issues with the latest commit on master in the test coverage. I think is has to do with the block module. Specifically the `raw_test.go` file. You can run `make test-coverage` to see these errors.

txpool/tx_object_test.go
MakisChristou: I don't remember the exact path of the code but its probably that one. The purpose was to get more coverage in areas that I could not cover otherwise with the default timestamp.

libotony: I would suggest two approaches:
1. Use the genesis builder to build a minimal genesis, like [this](https://github.com/vechain/thor/blob/master/consensus/consensus_test.go#L56)
2. Add a new block with the current time, add it to the repository, and set it with best would make loops inside txpool start working.

txpool/blocklist_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"net/http/httptest"
+	"os"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vechain/thor/v2/thor"
+)
+
+func SetupFile(t *testing.T, testFilePath string, dummyData string) {
```
libotony, line 16: How about storing the test file in the OS's [temporary](https://pkg.go.dev/os#CreateTemp) dir?

txpool/blocklist_test.go
```diff
@@ -0,0 +1,169 @@
...
+	err := os.WriteFile(testFilePath, []byte(dummyData), 0644)
+	if err != nil {
+		t.Fatalf("Failed to create test file: %s", err)
+	}
+}
+
+func TestLoad(t *testing.T) {
+
```
libotony, line 24: Is this new empty line on purpose?  I saw serval functions has a new empty line.

txpool/tx_object_test.go
libotony: @MakisChristou  Another approach https://github.com/vechain/thor/blob/master/genesis/customnet.go#L35, but I like to use the genesis builder for testing purposes.

PR #633, opened 2023-12-21
Docker security fix

darrenvechain:
We should use a specific env with protection rules when publishing to docker.

See documentation here: https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment

We will need to add the secrets to environment `docker-publish` and remove the secrets from actions


libotony:
Looks like the environment is already been set, so I need to move the secrets there

libotony (approved):

PR #636, opened 2024-01-02
Tx pool tests

MakisChristou:
@libotony It was easier for me to create a new PR from scratch. I added your suggestions refactoring the `NewDevnet` function to take a timestamp argument instead of duplicating it. I also used the `CreateTemp` function in our test cases as you suggested. Code coverage is above 80% in this PR.


txpool/tx_pool_test.go
```diff
@@ -40,6 +43,177 @@ func newPool(limit int, limitPerAccount int) *TxPool {
 		MaxLifetime:     time.Hour,
 	})
 }
+
+func newPoolWithParams(limit int, limitPerAccount int, BlocklistCacheFilePath string, BlocklistFetchURL string, timestamp uint64) *TxPool {
+	db := muxdb.NewMem()
+	gene := genesis.NewDevnet(timestamp)
```
libotony, line 49:
```suggestion
	gene := new(genesis.Builder).
		GasLimit(thor.InitialGasLimit).
		Timestamp(timestamp).
		State(func(state *state.State) error {
			bal, _ := new(big.Int).SetString("1000000000000000000000000000", 10)
			for _, acc := range genesis.DevAccounts() {
				state.SetBalance(acc.Address, bal)
				state.SetEnergy(acc.Address, bal, timestamp)
			}
			return nil
		})
```
This one if fine for this test, so I would suggest no need to change on `genesis.NewDevnet`

txpool/tx_pool_test.go
```diff
@@ -40,6 +43,177 @@ func newPool(limit int, limitPerAccount int) *TxPool {
...
+			w.WriteHeader(http.StatusNotModified)
+			return
+		}
+		fmt.Fprint(w, data)
+	}))
+	return server
+}
+
+func TestNewCloseWithServer(t *testing.T) {
+
```
libotony, line 78:
```suggestion
```

txpool/blocklist_test.go
```diff
@@ -0,0 +1,165 @@
...
+	if err != nil {
+		t.Errorf("Load failed: %s", err)
+	}
+
+	// Clean up: delete the temporary file.
+	os.Remove(testFilePath)
+}
+
+func TestLoadWithError(t *testing.T) {
+
```
libotony, line 49:
```suggestion
```

MakisChristou:
By the way regarding the empty lines its probably cause by `go fmt` since its running every time I save on VS Code

libotony:
> By the way regarding the empty lines its probably cause by `go fmt` since its running every time I save on VS Code

I am also using vscode plus go language plugin, it does not add new lines of every new function but accepts one empty line.

libotony:
BTW, the tests are failing since `math/big` is missing in txpool_test.go

libotony (approved):

PR #650, opened 2024-01-10
Add genesis tests

MakisChristou:
This PR addresses this [issue](https://github.com/vechainfoundation/protocol-board-repo/issues/22), by including test cases that increase the genesis coverage from an initial

builder   -> 80%
customnet -> 0%
devnet    -> 0%
genesis   -> 60%
mainnet   -> 0%
testnet   -> 70.6%

to 

builder   -> 90%
customnet -> 75%
devnet    -> 75.7%
genesis   -> 73.3%
mainnet   -> 76.4%
testnet   -> 70.6%

The lack of 80% coverage across all files is due to the fact that some parts of the code are unreachable. For example see the below snippet. To my understanding we are setting some global hardcoded values to the state and we have some error checking code. Since the value is constant and valid, the error case is never executed thus showing up as uncovered code. 

```golang
// setup builtin contracts
if err := state.SetCode(builtin.Authority.Address, builtin.Authority.RuntimeBytecodes()); err != nil {
        return err
}
```

Let me know if I am missing something or if this is indeed coverable code.


genesis/devnet.go
```diff
@@ -123,3 +123,71 @@ func NewDevnet() *Genesis {
 
 	return &Genesis{builder, id, "devnet"}
 }
+
+// NewDevnet create genesis for solo mode.
+func NewDevnetCustomTimestamp(timestamp uint64) *Genesis {
```
libotony, line 128: This function is unused.

genesis/genesis_test.go
```diff
@@ -22,6 +22,12 @@ func TestTestnetGenesis(t *testing.T) {
 	b0, _, _, err := gene.Build(state.NewStater(db))
 	assert.Nil(t, err)
 
+	id := gene.ID()
+	name := gene.Name()
+
+	assert.Equal(t, id, thor.Bytes32{0x0, 0x0, 0x0, 0x0, 0xb, 0x2b, 0xce, 0x3c, 0x70, 0xbc, 0x64, 0x9a, 0x2, 0x74, 0x9e, 0x86, 0x87, 0x72, 0x1b, 0x9, 0xed, 0x2e, 0x15, 0x99, 0x7f, 0x46, 0x65, 0x36, 0xb2, 0xb, 0xb1, 0x27})
```
libotony, line 28: `thor.MustParseBytes32("0x000000000b2bce3c70bc649a02749e8687721b09ed2e15997f466536b20bb127")` might be better for reading

PR #652, opened 2024-01-11
Add builtin tests

MakisChristou:
This PR addresses this [issue](https://github.com/vechainfoundation/protocol-board-repo/issues/31). Tests have been added for `energy.go` to increase coverage from 49.15 to 79.7%. The reason for not having above 80% coverage is uncoverable code. For example see the snippet below:

```golang
func (e *Energy) getInitialSupply() (init initialSupply, err error) {
	err = e.state.DecodeStorage(e.addr, initialSupplyKey, func(raw []byte) error {
		if len(raw) == 0 {
			init = initialSupply{&big.Int{}, &big.Int{}, 0}
			return nil
		}
		return rlp.DecodeBytes(raw, &init)
	})
	return
}
```

In the snippet above we want to cover the path inside the `if len(raw) == 0` path. The issue is that this is only (to my understanding) affected by the `initialSupplyKey` which is hardcoded and cannot be passed as argument. Other uncovered code relies on this snippet to return error thus leaving enough uncovered code to not reach 80%. 

Another way we could reach the code is to allow the `New` function in `state.go` to take the function we pass in `stackedMap.New()` as argument. Thus we can pass a function that always errors.

### DIFF
--- a/builtin/energy/energy_test.go
+++ b/builtin/energy/energy_test.go
@@ -35,11 +35,83 @@ func TestEnergy(t *testing.T) {
 		{M(eng.Get(acc)), M(big.NewInt(10), nil)},
 		{M(eng.Sub(acc, big.NewInt(5))), M(true, nil)},
 		{M(eng.Sub(acc, big.NewInt(6))), M(false, nil)},
+		{eng.Add(acc, big.NewInt(0)), nil},
+		{M(eng.Sub(acc, big.NewInt(0))), M(true, nil)},
 	}
 
 	for _, tt := range tests {
 		assert.Equal(t, tt.expected, tt.ret)
 	}
+}
+
+func TestInitialSupply(t *testing.T) {
+	db := muxdb.NewMem()
+	st := state.New(db, thor.Bytes32{}, 0, 0, 0)
+
+	eng := New(thor.BytesToAddress([]byte("eng")), st, 0)
+
+	eng.SetInitialSupply(big.NewInt(123), big.NewInt(456))
+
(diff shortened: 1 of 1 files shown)


libotony:
> This PR addresses this [issue](https://github.com/vechainfoundation/protocol-board-repo/issues/31). Tests have been added for `energy.go` to increase coverage from 49.15 to 79.7%. The reason for not having above 80% coverage is uncoverable code. For example see the snippet below:
> 
> ```go
> func (e *Energy) getInitialSupply() (init initialSupply, err error) {
> 	err = e.state.DecodeStorage(e.addr, initialSupplyKey, func(raw []byte) error {
> 		if len(raw) == 0 {
> 			init = initialSupply{&big.Int{}, &big.Int{}, 0}
> 			return nil
> 		}
> 		return rlp.DecodeBytes(raw, &init)
> 	})
> 	return
> }
> ```
> 
> In the snippet above we want to cover the path inside the `if len(raw) == 0` path. The issue is that this is only (to my understanding) affected by the `initialSupplyKey` which is hardcoded and cannot be passed as argument. Other uncovered code relies on this snippet to return error thus leaving enough uncovered code to not reach 80%.
> 
> Another way we could reach the code is to allow the `New` function in `state.go` to take the function we pass in `stackedMap.New()` as argument. Thus we can pass a function that always errors.

Check https://github.com/vechain/thor/pull/652/commits/1d3288af0fad70cac293cf880e65b95806d52cc0

libotony (approved):

PR #653, opened 2024-01-12
Add genesis tests to increase coverage

MakisChristou:
This PR addresses this [issue](https://github.com/vechainfoundation/protocol-board-repo/issues/22), by including test cases that increase the genesis coverage from an initial

builder   -> 80%
customnet -> 0%
devnet    -> 0%
genesis   -> 60%
mainnet   -> 0%
testnet   -> 70.6%

to 

builder   -> 90%
customnet -> 75%
devnet    -> 75.7%
genesis   -> 73.3%
mainnet   -> 76.4%
testnet   -> 70.6%

The lack of 80% coverage across all files is due to the fact that some parts of the code are unreachable. For example see the below snippet. To my understanding we are setting some global hardcoded values to the state and we have some error checking code. Since the value is constant and valid, the error case is never executed thus showing up as uncovered code. 

```golang
// setup builtin contracts
if err := state.SetCode(builtin.Authority.Address, builtin.Authority.RuntimeBytecodes()); err != nil {
        return err
}
```

Let me know if I am missing something or if this is indeed coverable code.

@libotony Regarding the unused function its actually used to increase coverage in `TestNewDevnetCustomTimestamp`

### DIFF
--- a/genesis/customnet_test.go
+++ b/genesis/customnet_test.go
@@ -1,0 +1,148 @@
+package genesis_test
+
+import (
+	"math"
+	"math/big"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vechain/thor/v2/genesis"
+	"github.com/vechain/thor/v2/thor"
+)
+
+func CustomNetWithParams(t *testing.T, executor genesis.Executor, baseGasPrice genesis.HexOrDecimal256, rewardRatio genesis.HexOrDecimal256, proposerEndorsement genesis.HexOrDecimal256) genesis.CustomGenesis {
+	var defaultFC = thor.ForkConfig{
+		VIP191:    math.MaxUint32,
+		ETH_CONST: math.MaxUint32,
+		BLOCKLIST: math.MaxUint32,
+		ETH_IST:   math.MaxUint32,
+		VIP214:    math.MaxUint32,
+		FINALITY:  0,
+	}
+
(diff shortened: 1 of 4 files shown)


libotony:
> @libotony Regarding the unused function its actually used to increase coverage in `TestNewDevnetCustomTimestamp`

If I remember correctly, the function `NewDevnetCustomTimestamp` was introduced for txpool tests in #636 and we didn't use it at the end, so I am suggesting remove both `NewDevnetCustomTimestamp` and `TestNewDevnetCustomTimestamp`, since `TestNewDevnetCustomTimestamp` is not used in the code base.

libotony:
> The lack of 80% coverage across all files is due to the fact that some parts of the code are unreachable. For example see the below snippet. To my understanding we are setting some global hardcoded values to the state and we have some error checking code. Since the value is constant and valid, the error case is never executed thus showing up as uncovered code.
> 
> ```go
> // setup builtin contracts
> if err := state.SetCode(builtin.Authority.Address, builtin.Authority.RuntimeBytecodes()); err != nil {
>         return err
> }
> ```

I would consider this acceptable, otherwise you might need to make a customized `stateDB` implementation that always throws error for  `setCode` @nwbrettski we can discuss this during our standups.

nwbrettski:
I agree @libotony this is acceptable. It is an improvement. We can always take another look to improve further at a later date if required.

libotony (approved):

PR #90, opened 2018-08-07
Added TheVechain.com block explorer to the list.

thevechain:
Also fixed some english.

PR #661, opened 2024-01-29
Fixes #90. Adds proper handling and tests when block id does not exists

ahasikos:
# Description

When a particular block id is requested from the API and does not exists we return a Status Not Found(404) Error whereas previously we were returning a Status OK with a null body.

Fixes https://github.com/vechainfoundation/protocol-board-repo/issues/90

# How Has This Been Tested?

Added a new test in `blocks_test.go`

`make test`

# Checklist:


darrenvechain (changes requested):
This PR introduces a breaking change. As a best practice, it's crucial to ensure that existing endpoints remain backward compatible.

We would need a `v2` endpoint to throw a 404 for blocks not found

PR #676, opened 2024-03-06
test: add test coverage for api/utils package

paologalligit:
# Description

This commit adds a test coverage suite for the `api/utils` package.

# Checklist:

### DIFF
--- a/api/utils/http_test.go
+++ b/api/utils/http_test.go
@@ -1,0 +1,126 @@
+// Copyright (c) 2018 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package utils
+
+import (
+	"bytes"
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


libotony (approved):

api/utils/http_test.go
```diff
@@ -0,0 +1,126 @@
+// Copyright (c) 2018 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package utils
```
libotony, line 6: Does utils_test work?

libotony (changes requested):
Dismiss my previous approval reivew.

api/utils/http_test.go
paologalligit: Yes, it does. In this case I'm only using public functions from utils package. I was reading this [page](https://go.dev/doc/tutorial/add-a-test) and here it suggests to put tests in the same package as the implementation. I've realized there are different ways to approach this, do we have any convention we're following?

libotony: I was under the impression that the package with the suffix `_test` would not be compiled into the official binary so when I was doing tests, if I don't use the private functions I would use `xxx_test` as the package name.

libotony (approved):

PR #685, opened 2024-03-28
Enable e2e tests

darrenvechain:
# Description

This PR enables E2E Testing

## Type of change

# How Has This Been Tested?

See the GitHub action result

# Checklist:

### DIFF
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -43,8 +43,8 @@ jobs:
         uses: actions/checkout@v4
         with:
           repository: vechain/thor-e2e-tests
-          # TODO: Should we use tags here? eg 'v2.1.0'
-          ref: main
+          # https://github.com/vechain/thor-e2e-tests/tree/2cb22d804bb3cdf075917dbece42a182d42d7486
+          ref: 2cb22d8
 
       - name: Download artifact
         uses: actions/download-artifact@v4
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -44,7 +44,7 @@ jobs:
         with:
           repository: vechain/thor-e2e-tests
           # https://github.com/vechain/thor-e2e-tests/tree/2cb22d804bb3cdf075917dbece42a182d42d7486
-          ref: 2cb22d8
+          ref: 2cb22d804bb3cdf075917dbece42a182d42d7486
 
       - name: Download artifact
         uses: actions/download-artifact@v4
--- a/.github/workflows/test-e2e.yaml
+++ b/.github/workflows/test-e2e.yaml
@@ -1,0 +1,81 @@
+name: E2E Tests
+
+on:
+  pull_request:
+    branches:
+      - master
+  push:
+    branches:
+      - master
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,76 @@
...
+      THOR_IMAGE: vechain/thor:${{ github.sha }}
+    name: Run Tests
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: vechain/thor-e2e-tests
+          # TODO: Should we use tags here? eg 'v2.1.0'
+          ref: main
```
claytonneal, line 47: Perhaps need to think about this, wdyt @leszek-vechain

darrenvechain: I think `main` probably isn't the best, as we could create new tests that start breaking the thor CI.

IMO a commit SHA could be a good option for this. Once you write code / break existing tests, you should modify the e2e, create a PR, merge it, and then update the SHA in the GHA.

The same approach would have to be taken in the context of submodules too, since you would have to update e2e and then update the submodule reference.

So I could set the value to [fba3138](https://github.com/vechain/thor-e2e-tests/commit/fba3138cc140c8f734f68bdd4bbd6d27b670d7e5), which is the latest commit

leszek-vechain: need to think about it - it seems like it will be quite a lot of work, maybe we can just move e2e to thor repo, in that case, you change code, you change tests in some commit

.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,76 @@
...
+      THOR_IMAGE: vechain/thor:${{ github.sha }}
+    name: Run Tests
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: vechain/thor-e2e-tests
+          # https://github.com/vechain/thor-e2e-tests/tree/2cb22d804bb3cdf075917dbece42a182d42d7486
+          ref: 2cb22d8
```
leszek-vechain, line 47: so instead of targeting main we can target certain commit for tests (or even branch), is there anyway we can automate it instead of using hardcoded value ? something like specifying it before running workflow

darrenvechain: Hhmm, how do you mean? Where would we specify it?

leszek-vechain: thats a good question, which I don't have answer too :)

leszek-vechain (approved):

.github/workflows/test-e2e.yaml
```diff
@@ -0,0 +1,76 @@
...
+    env:
+      THOR_IMAGE: vechain/thor:${{ github.sha }}
+    name: Run Tests
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          repository: vechain/thor-e2e-tests
+          # https://github.com/vechain/thor-e2e-tests/tree/2cb22d804bb3cdf075917dbece42a182d42d7486
```
otherview, line 46: What is this comment ?

PR #686, opened 2024-03-29
Debug test

paologalligit:
# Description

Added tests for the `api/debug` packages. With this test suite its coverage is 86.7%.

# Checklist:

### DIFF
--- a/api/debug/debug.go
+++ b/api/debug/debug.go
@@ -198,11 +198,6 @@ func (d *Debug) handleTraceCall(w http.ResponseWriter, req *http.Request) error 
 	return utils.WriteJSON(w, res)
 }
 
-type extractTracerOption struct {
-	Name   string
-	Config json.RawMessage
-}
-
 func (d *Debug) createTracer(name string, config json.RawMessage) (tracers.Tracer, error) {
 	if name == "" {
 		return logger.NewStructLogger(config)
--- a/api/debug/debug.go
+++ b/api/debug/debug.go
@@ -153,8 +153,8 @@ func (d *Debug) handleTraceClause(w http.ResponseWriter, req *http.Request) erro
 		return utils.BadRequest(errors.WithMessage(err, "body"))
 	}
 
-	var tracer tracers.Tracer
-	if err := extractTracer(&extractTracerOption{Name: opt.Name, Config: opt.Config}, &tracer); err != nil {
+	tracer, err := d.createTracer(opt.Name, opt.Config)
+	if err != nil {
 		return utils.Forbidden(err)
 	}
 
@@ -180,8 +180,8 @@ func (d *Debug) handleTraceCall(w http.ResponseWriter, req *http.Request) error 
 		return err
 	}
 
-	var tracer tracers.Tracer
-	if err := extractTracer(&extractTracerOption{Name: opt.Name, Config: opt.Config}, &tracer); err != nil {
+	tracer, err := d.createTracer(opt.Name, opt.Config)
+	if err != nil {
 		return utils.Forbidden(err)
 	}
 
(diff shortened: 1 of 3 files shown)


tracers/tracers.go
```diff
@@ -82,7 +87,14 @@ func (d *directory) RegisterJSEval(f jsCtorFn) {
 // New returns a new instance of a tracer, by iterating through the
 // registered lookups. Name is either name of an existing tracer
 // or an arbitrary JS code.
-func (d *directory) New(name string, cfg json.RawMessage, allowCustom bool) (Tracer, error) {
+func (d *directory) New(name string, cfg json.RawMessage, allowCustom bool) (tr Tracer, err error) {
+	defer func() {
```
libotony, line 91: Not necessary

api/debug/debug.go
```diff
@@ -215,6 +198,28 @@ func (d *Debug) handleTraceCall(w http.ResponseWriter, req *http.Request) error
 	return utils.WriteJSON(w, res)
 }
 
+type extractTracerOption struct {
+	Name   string
+	Config json.RawMessage
+}
+
+func extractTracer(opt *extractTracerOption, tracer *tracers.Tracer) error {
```
libotony, line 206: How about `(d *Debug) createTracer(name string, config json.RawMessage)(tracers.Tracer, error)` ?

libotony (changes requested):

tracers/tracers.go
paologalligit: This is to cover the case the body request has a specific tracer name (in the case `opt.Name` is not an empty string we allow custom tracers), but it doesn't exist. In that case this [line](https://github.com/vechain/thor/blob/8a65c7f6bcc182bc889b982a09aaf934866a3021/tracers/tracers.go#L108) will panic

api/debug/debug.go
paologalligit: Are there any advantages in making it a struct method? Because when I created it I thought about it as a utility function to extract the tracer from config, and not as a part of the debug struct, but I might be missing something

tracers/tracers.go
libotony: Sorry I am not following, could you give a case that the specific line will panic?

paologalligit: calling the `debug/tracers` endpoint with a payload like in [this](https://github.com/vechain/thor/blob/8a65c7f6bcc182bc889b982a09aaf934866a3021/api/debug/debug_test.go#L165) test will panic, but actually calling the endpoint against a real node instance returns an error. Not sure why this happens, probably the `jsEval` method hides some context that is missing during a test run. I'll remove it, then

libotony: This is a tricky one, panic is caused by not loading js tracer engine, you need to add `_"github.com/vechain/thor/v2/tracers/js"` to the import section

api/debug/debug.go
libotony: One reason is keep the same style of all API packages, another one is sharing the option `d. allowCustomTracer `

paologalligit: got it, done

tracers/tracers.go
libotony: Looks like you have removed the `nonExistTracer` test in https://github.com/vechain/thor/pull/686/commits/adcfa604b1666b1c0aa05e8e29efa9c7fd7777c5, do you have a plan adding it back or just leave it as is?

paologalligit: I think I'll add it back, it doesn't deserve to be deleted after all 😆

tracers/tracers.go
```diff
@@ -27,6 +27,11 @@ import (
 	"github.com/vechain/thor/v2/vm"
 )
 
+var (
+	// ErrUnsupportedTracer is returned when a tracer cannot be created.
+	ErrUnsupportedTracer = errors.New("cannot create custom tracer")
+)
+
```
libotony, line 34: No longer needed.

PR #688, opened 2024-04-01
Adds allowed-peers Flag

otherview:
# Description

Adds a new flag that limits the connections of a node. 
This features help and allows for a few features like bootstrapping to be limited to a single node (good for metrics), DMZ's and traffic-shapping in general is now easier.

Flags accept a list of enodes :
```
--allowed-peers enode://797fdd968592ca3b59a143f1aa2f152913499d4bb469f2bd5b62dfb1257707b4cb0686563fe144ee2088b1cc4f174bd72df51dbeb7ec1c5b6a8d8599c756f38b@107.150.112.22:55555 enode://3eae6740af6180bb015309f7a07ff7405d6f1f9f1e5a9f2fabbd36b0c00b862521e63ff23573ffdb9035f2237c26513cb9f02454f9ada993e60b99ffc187bb54@107.150.112.21:55555
```

Fixes # (issue)

## Type of change

# How Has This Been Tested?
- [x] Log testing, making sure that no other peer had connected


claytonneal:
heya, do we need up update any README or docs.vechain.org for this?

p2psrv/server.go
```diff
@@ -47,6 +48,18 @@ func New(opts *Options) *Server {
...
+	if len(opts.ConnectOnlyNodes) > 0 {
+		knownNodes = cache.NewPrioCache(5)
+		discoveredNodes = cache.NewRandCache(128)
+		for _, connectOnlyNode := range opts.ConnectOnlyNodes {
+			knownNodes.Set(connectOnlyNode.ID, connectOnlyNode, 0)
+			discoveredNodes.Set(connectOnlyNode.ID, connectOnlyNode)
+			connectOnlyNodes.Add(connectOnlyNode)
+		}
+	}
+
```
leszek-vechain, line 62: is there anyway we can test this on unit/integration level ?

otherview: Yup! Added a test in the New() method. I would like to add more tests, but I believe without refactoring the code it wouldn't be a very efficient test code.

leszek-vechain (dismissed):

libotony:
We already have a `--bootnode` flag that does part of the feature but does not limit other connections. I would consider adding a new flag `--no-discovery` that disables the discovery feature.

otherview:
Myself and @libotony did a pair-prog on this, waiting on one more review.

cmd/thor/flags.go
```diff
@@ -142,4 +142,10 @@ var (
 		Value: 16,
 		Usage: "set tx limit per account in pool",
 	}
+
+	allowedPeersOnlyFlag = cli.StringSliceFlag{
```
darrenvechain, line 146: Should this flag be hidden? If not, I would be inclined to remove the examples, as they may clog the `help` command.

Also I think `allowed-peers` is enough, it doesn't need the `-only` IMO

darrenvechain (approved):
Non-blocking thoughts around the flag, but looks good

p2psrv/server.go
```diff
@@ -47,6 +48,18 @@ func New(opts *Options) *Server {
 		discoveredNodes.Set(node.ID, node)
 	}
 
+	// AllowedPeers doesn't connect to other sourced nodes
+	allowedOnlyNodes := newNodeMap()
+	if len(opts.AllowedPeers) > 0 {
```
paologalligit, line 53: What about having this condition at the start so we don't re-write the `knownNodes` and `discoveredNodes`?
```
allowedOnlyNodes := newNodeMap()
knownNodes := cache.NewPrioCache(5)
discoveredNodes := cache.NewRandCache(128)

if len(opts.AllowedPeers) > 0 {
        for _, connectOnlyNode := range opts.AllowedPeers {
	        knownNodes.Set(connectOnlyNode.ID, connectOnlyNode, 0)
	        discoveredNodes.Set(connectOnlyNode.ID, connectOnlyNode)
	        allowedOnlyNodes.Add(connectOnlyNode)
        }
} else {
	for _, node := range opts.KnownNodes {
		knownNodes.Set(node.ID, node, 0)
		discoveredNodes.Set(node.ID, node)
	}
}
```

cmd/thor/utils.go
```diff
@@ -440,6 +450,8 @@ func newP2PComm(ctx *cli.Context, repo *chain.Repository, txPool *txpool.TxPool,
 		BootstrapNodes:  fallbackBootstrapNodes,
 		RemoteBootstrap: remoteBootstrapList,
 		NAT:             nat,
+		AllowedPeers:    allowedPeers,
+		NoDiscovery:     len(ctx.StringSlice(allowedPeersFlag.Name)) > 0,
```
paologalligit, line 454: Am I missing something or this `NoDiscovery` here is the same passed to [new](https://github.com/vechain/thor/blob/a82d2dbfe53c3568fabc1313bb7841179d0db2ac/cmd/thor/utils.go#L485) but then is always hardcoded as true [here](https://github.com/vechain/thor/blob/a82d2dbfe53c3568fabc1313bb7841179d0db2ac/p2psrv/server.go#L70)?

otherview: We only [store/use](https://github.com/vechain/thor/blob/a82d2dbfe53c3568fabc1313bb7841179d0db2ac/p2psrv/server.go#L64) the `opts` [in our business logic to discover new peers ](https://github.com/vechain/thor/blob/a82d2dbfe53c3568fabc1313bb7841179d0db2ac/p2psrv/server.go#L122)

But you're right it does look confusing, I'll add a comment that the p2p server default discovery mechanism is not used per default.

p2psrv/server.go
otherview: I can see that if this was a high cycled operation it would have an impact, but I think it might be a bit of a style pref.
Do you strongly feel about this ?

paologalligit: No, if you prefer it this way, it's fine :)

paologalligit:
> Myself and @libotony did a pair-prog on this, due to flag behaviours we ended up agreeing on `allow-peers-only` as a flag that only allows connectivity to certain peers.
> 
> The TLDR is that --no-discovery + --bootnodes in certain cases can still allow connections to other nodes.

About the last line, is it because this is only an outgoing connection restricted list and so it doesn't imply that another node could start an incoming connection with the node (even if not in the allow-peers flag)?

cmd/thor/utils.go
```diff
@@ -438,6 +438,16 @@ func newP2PComm(ctx *cli.Context, repo *chain.Repository, txPool *txpool.TxPool,
 		return nil, errors.Wrap(err, "parse -nat flag")
 	}
 
+	// recreate the node from the enode string slice
+	var allowedPeers p2psrv.Nodes
```
libotony, line 442: How about leveraging the existing `KnownNodes`?

cmd/thor/flags.go
```diff
@@ -147,4 +147,11 @@ var (
 		Name:  "genesis",
 		Usage: "path to genesis file, if not set, the default devnet genesis will be used",
 	}
+
+	allowedPeersFlag = cli.StringSliceFlag{
```
libotony, line 151: How about align with `bootnode` flag, using `comma separated list`?

otherview: Done ✅

cmd/thor/utils.go
```diff
@@ -424,56 +424,78 @@ type p2pComm struct {
...
-		BootstrapNodes:  fallbackBootstrapNodes,
-		RemoteBootstrap: remoteBootstrapList,
-		NAT:             nat,
-	}
+	var knowNodes p2psrv.Nodes
+	var restrictedPeerListEnabled bool
+	remoteBootstrapURL := remoteBootstrapList
 
-	peersCachePath := filepath.Join(instanceDir, "peers.cache")
+	// restricted peer list mode - no other peers are loaded
```
otherview, line 450: Reframed the existing behaviour to be clear.
- If there is a restricted list ONLY load that list into the KnownNodes
- Otherwise : 
  - Load default or flag-supplied boot nodes, and append those to any cache-stored nodes

p2psrv/options.go
```diff
@@ -40,11 +40,6 @@ type Options struct {
 
 	KnownNodes Nodes
 
-	// BootstrapNodes are used to establish connectivity
-	// with the rest of the network using the V5 discovery
-	// protocol.
-	BootstrapNodes Nodes
-
```
otherview, line 47: This seems to be redundant given that the BootstrapNodes = CachedNodes + Bootstrap nodes that was set in the utils files

p2psrv/server.go
```diff
@@ -197,9 +197,6 @@ func (s *Server) listenDiscV5() (err error) {
 		}
 	}()
 
-	for _, node := range s.opts.BootstrapNodes {
-		s.bootstrapNodes = append(s.bootstrapNodes, discv5.NewNode(discv5.NodeID(node.ID), node.IP, node.UDP, node.TCP))
-	}
```
otherview, line 202: Already done in the utils file - removes complexity

libotony (approved):

paologalligit (approved):

PR #723, opened 2024-04-29
Add custom thor solo block interval

MakisChristou:
Tackles the following [issue](https://github.com/orgs/vechain/projects/22/views/1?filterQuery=label%3A%22App%3A+Thor%22%2C%22App%3A+VIP%22%2C%22App%3A+Thorest%22++solo&pane=issue&itemId=53228366). In this PR I basically added the option to use a command-line argument specifically for solo to allow the developer to choose a custom block interval with a minimum value of 1. I tested it and it looks like its working fine for 1 second block intervals as well as larger intervals like 100 seconds.

### DIFF
--- a/cmd/thor/flags.go
+++ b/cmd/thor/flags.go
@@ -131,7 +131,7 @@ var (
 	blockInterval = cli.IntFlag{
 		Name:  "block-interval",
 		Value: 10,
-		Usage: "choose a custom block interval for solo mode",
+		Usage: "choose a custom block interval for solo mode (seconds)",
 	}
 	persistFlag = cli.BoolFlag{
 		Name:  "persist",
--- a/cmd/thor/flags.go
+++ b/cmd/thor/flags.go
@@ -128,6 +128,11 @@ var (
 		Name:  "on-demand",
 		Usage: "create new block when there is pending transaction",
 	}
+	blockInterval = cli.IntFlag{
+		Name:  "block-interval",
+		Value: 10,
+		Usage: "choose a custom block interval for solo mode",
+	}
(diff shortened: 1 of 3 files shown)


cmd/thor/flags.go
```diff
@@ -128,6 +128,11 @@ var (
 		Name:  "on-demand",
 		Usage: "create new block when there is pending transaction",
 	}
+	blockInterval = cli.IntFlag{
+		Name:  "interval",
```
darrenvechain, line 132: `block-interval` maybe? I think interval is too generic

MakisChristou: What about now?

paologalligit (approved):
LGTM

cmd/thor/flags.go
```diff
@@ -128,6 +128,11 @@ var (
 		Name:  "on-demand",
 		Usage: "create new block when there is pending transaction",
 	}
+	blockInterval = cli.IntFlag{
+		Name:  "block-interval",
+		Value: 10,
+		Usage: "choose a custom block interval for solo mode",
```
libotony, line 134: Should mention it was in `seconds`?

cmd/thor/flags.go
```diff
@@ -128,6 +128,11 @@ var (
 		Name:  "on-demand",
 		Usage: "create new block when there is pending transaction",
 	}
+	blockInterval = cli.IntFlag{
+		Name:  "block-interval",
+		Value: 10,
```
libotony, line 133: The value should be checked, shouldn't be 0.

PR #724, opened 2024-04-29
Cleanup p2p thor + add tests

otherview:
# Description

Connected with https://github.com/vechain/protocol-board-repo/issues/165 .
It was hard to follow the expected behaviour when combining (`allowed-peers`, `bootnodes`, and stored cached peers). 
This was mainly due to the use of flag value retrieval, directory setup, and a few other operations being mixed with the flag operation behaviour.
The PR decouples the setup of data and directories from the flag combination logic.
It was also very hard to test. This PR does a bit of refactoring around the `p2pcom` and makes it more testable.
The logic now works as follows:
- If no flag is set, load `fallbackbootnodes` and try to load `cachedPeers`, append these two.
- if `bootnodes` is set, ignore `fallbackbootnodes` and try to load `cachedPeers`, append these two.
- If `allowedPeers` is set, ignore `bootnodes`, `fallbackbootnodes` and `cachedPeers`.
- Never allow for nodes to be duplicated when appending.

This PR also introduces a `test/datagen` package that makes it easier to generate randomized data.

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:


p2psrv/server.go
```diff
@@ -197,9 +197,6 @@ func (s *Server) listenDiscV5() (err error) {
 		}
 	}()
 
-	for _, node := range s.opts.BootstrapNodes {
-		s.bootstrapNodes = append(s.bootstrapNodes, discv5.NewNode(discv5.NodeID(node.ID), node.IP, node.UDP, node.TCP))
-	}
```
otherview, line 202: `s.opts.BootstrapNodes` is only used here.

test/datagen/numbers.go
```diff
@@ -0,0 +1,13 @@
+package datagen
```
otherview, line 1: Helps to standartize the generation of random data.

p2psrv/options.go
```diff
@@ -40,11 +40,6 @@ type Options struct {
 
 	KnownNodes Nodes
 
-	// BootstrapNodes are used to establish connectivity
-	// with the rest of the network using the V5 discovery
-	// protocol.
-	BootstrapNodes Nodes
-
```
otherview, line 47: This comment is outdated BootstrapNodes are only ever used to pass nodes from the previous `p2pcomm` to the p2psrv, at which point they would be appended with the `KnownNodes` into the `s.bootstrapNodes`

cmd/thor/utils.go
```diff
@@ -652,8 +593,6 @@ func parseNodeList(list string) ([]*discover.Node, error) {
 		}
 		nodes = append(nodes, node)
 	}
-	if len(nodes) == 0 {
-		return nil, errors.New("empty node list")
-	}
```
otherview, line 657: The expectancy of a list with items is now removed so we can parse empty lists

p2psrv/options.go
otherview: Before:
- BootstrapNodes + KnownNodes were used to discover new nodes
- KnownNodes were used to connect and discover new nodes + p2p communication

Now:
- KnowNodes are used to discover new nodes  + p2p communication

This cleans up the process. The idea behind this is that it should be the serving node responsibility to serve JUST discovery or discovery + p2p. It should not be the client's responsibility, because this line of action does not protect discovery nodes.

libotony: I might be against this idea, `BootstrapNodes` and `KnownNodes` serves their different purpose, shouldn't mix them together. 

`BootStrapNodes` is for discovery protocol.
`KnownNodes` is known(connected or node operator supplied) peer-to-peer nodes.

otherview: Yeah I understand that they serve different purposes, I just think that the idea behind it should be cleaner or simpler.

Why can't the `Bootstrap nodes` also be connected over p2p ? 
I guess in my head, it should be the receiving node responsibility to accept or deny the protocols it talks in.

otherview: This was bugging me, took me a bit to figure out why, thanks for the patience and the review.

Fundamentally, I was adding a new feature in the middle of a clean-up/refactor - which I believe can arguably be a good simplification - but not the point of the PR.
The point of the PR was to make the flow of the usage of flags easy to follow and to have a testable p2p package.
I've :
- Reverted the removal of `BootstrapNodes` (no new feature)
- Renamed `BootstrapNodes` to DiscoveryNodes; `RemoteBootstrap` to `RemoteDiscoveryList` and `fallbackBootstrapNodes` to `fallbackDiscoveryNodes`
- Updated the tests

The functionality remains the same, but this is fine-tuning the names. It feels to me that DiscoveryNodes is a less prone to confusion name for the initial default connection nodes exclusively used for peer discovery (and never used for p2p).

cmd/thor/utils.go
```diff
@@ -437,98 +432,44 @@ func newP2PComm(ctx *cli.Context, repo *chain.Repository, txPool *txpool.TxPool,
...
 
-	opts := &p2psrv.Options{
-		Name:            common.MakeName("thor", fullVersion()),
-		PrivateKey:      key,
-		MaxPeers:        ctx.Int(maxPeersFlag.Name),
-		ListenAddr:      fmt.Sprintf(":%v", ctx.Int(p2pPortFlag.Name)),
-		BootstrapNodes:  fallbackBootstrapNodes,
-		RemoteBootstrap: remoteBootstrapList,
-		NAT:             nat,
+	allowedPeers, err := parseNodeList(strings.TrimSpace(ctx.String(allowedPeersFlag.Name)))
```
libotony, line 441: `if flagAllowedPeers != "" {` required for the input validation and since the empty list check was removed, then required outside

cmd/thor/utils.go
```diff
@@ -416,14 +418,7 @@ func loadNodeMaster(ctx *cli.Context) (*node.Master, error) {
...
 
-type p2pComm struct {
-	comm           *comm.Communicator
-	p2pSrv         *p2psrv.Server
-	peersCachePath string
-	enode          string
-}
-
-func newP2PComm(ctx *cli.Context, repo *chain.Repository, txPool *txpool.TxPool, instanceDir string) (*p2pComm, error) {
+func newThorP2P(ctx *cli.Context, repo *chain.Repository, txPool *txpool.TxPool, instanceDir string) (*p2p.ThorP2P, error) {
```
libotony, line 421: `ThorP2P` sounds strange to me, what do you think about the old one `P2PComm`

cmd/thor/utils.go
libotony: An edge case for this might be --allowed-peers ",,," will give empty list, I think for this case explicitly declaring invalid input will make sense, wdyt?

leszek-vechain: worth checking this case

otherview: It's a good edge case, we're covered for it because of 
https://github.com/vechain/thor/pull/724/files#diff-cdfc7fee9b6dfccae42f1f5acc472fcaf660bfb5539d06c754a27502b12b8e00R590-R592

It will not try to parse a node because the string is "".

cmd/thor/utils.go
otherview: That kinda already exists right? Well not exactly p2pcomm but there is a comm package.. I found these confusing, to be honest that's why I try to split them.
I chose ThorP2P because it's the package that initiates all of thor's P2P needs (flags, discovery and comms). 
Not attached to it though, perhaps there another name that could work?

cmd/thor/p2p/p2p.go
```diff
@@ -0,0 +1,136 @@
...
+	// allowed peers flag will only allow p2psrv to connect to the designated peers
+	if len(allowedPeers) > 0 {
+		opts.NoDiscovery = true // disable discovery
+		opts.DiscoveryNodes = nil
+		opts.KnownNodes = allowedPeers
+	} else {
+		// boot nodes flag will overwrite the default bootstrap nodes and also disable remote bootstrap
+		if len(bootstrapNodes) > 0 {
+			opts.RemoteDiscoveryList = "" // disable remote bootstrap
+			opts.DiscoveryNodes = nil
```
libotony, line 69:
```suggestion
			opts.DiscoveryNodes = bootstrapNodes
```
It's good to have this line indication bootstrap nodes will also be used as discovery nodes

cmd/thor/p2p/p2p.go
```diff
@@ -0,0 +1,137 @@
...
+	if len(allowedPeers) > 0 {
+		opts.NoDiscovery = true // disable discovery
+		opts.DiscoveryNodes = nil
+		opts.KnownNodes = allowedPeers
+	} else {
+		// boot nodes flag will overwrite the default bootstrap nodes and also disable remote bootstrap
+		// bootstrap nodes will be connected for discovery and for p2p protocols
+		if len(bootstrapNodes) > 0 {
+			opts.RemoteDiscoveryList = ""    // disable remote bootstrap
+			opts.DiscoveryNodes = nil        // discovery nodes are unused given that known nodes are used for both node discovery and p2p protocol
```
otherview, line 70:
```suggestion
			opts.DiscoveryNodes = bootstrapNodes        // discovery nodes are unused given that known nodes are used for both node discovery and p2p protocol
```
```suggestion
			opts.DiscoveryNodes = nil        // discovery nodes are unused given that known nodes are used for both node discovery and p2p protocol
```

libotony (approved):

PR #727, opened 2024-04-30
Increase bandwidth coverage

MakisChristou:
Increases `bandwith.go` coverage to 84%. Relevant [issue](https://github.com/vechain/protocol-board-repo/issues/184).

### DIFF
--- a/block/header.go
+++ b/block/header.go
@@ -53,22 +53,6 @@ type headerBody struct {
 	Extension extension
 }
 
-func NewMockHeader() *Header {
-	return &Header{body: headerBody{
-		ParentID:        thor.Bytes32{1},
-		Timestamp:       1,
-		GasLimit:        100000,
-		Beneficiary:     thor.Address{1},
-		GasUsed:         11234,
-		TotalScore:      1,
-		TxsRootFeatures: txsRootFeatures{},
-		StateRoot:       thor.Bytes32{1},
-		ReceiptsRoot:    thor.Bytes32{1},
-		Signature:       []byte{1},
-		Extension:       extension{Alpha: []byte{1}, COM: true},
-	}}
-}
-
 // ParentID returns id of parent block.
 func (h *Header) ParentID() thor.Bytes32 {
 	return h.body.ParentID
--- a/block/header.go
+++ b/block/header.go
@@ -53,6 +53,22 @@ type headerBody struct {
 	Extension extension
 }
 
+func NewMockHeader() *Header {
+	return &Header{body: headerBody{
+		ParentID:        thor.Bytes32{1},
+		Timestamp:       1,
+		GasLimit:        100000,
+		Beneficiary:     thor.Address{1},
+		GasUsed:         11234,
+		TotalScore:      1,
+		TxsRootFeatures: txsRootFeatures{},
+		StateRoot:       thor.Bytes32{1},
+		ReceiptsRoot:    thor.Bytes32{1},
+		Signature:       []byte{1},
+		Extension:       extension{Alpha: []byte{1}, COM: true},
+	}}
+}
+
 // ParentID returns id of parent block.
 func (h *Header) ParentID() thor.Bytes32 {
 	return h.body.ParentID
--- a/cmd/thor/bandwidth/bandwidth_test.go
+++ b/cmd/thor/bandwidth/bandwidth_test.go
@@ -1,0 +1,64 @@
+package bandwidth
+
+import (
+	"crypto/rand"
+	"sync"
+	"testing"
+
+	"github.com/vechain/thor/v2/block"
+)
+
+func TestBandwidth(t *testing.T) {
+
+	bandwidth := Bandwidth{
+		value: 0,
+		lock:  sync.Mutex{},
+	}
+
+	val := bandwidth.Value()
+
+	if val != 0 {
+		t.Errorf("Expected 0, got %d", val)
+	}
+}
+
+func GetMockHeader(t *testing.T) *block.Header {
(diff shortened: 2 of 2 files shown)


block/header.go
```diff
@@ -53,6 +53,22 @@ type headerBody struct {
 	Extension extension
 }
 
+func NewMockHeader() *Header {
```
paologalligit, line 56: This function looks like something to be used only during testing. I would move it to `bandwidth_test.go` and make it private. To build the header I would take advantage of the block.Builder and extract the header. Something like
```
block := new(block.Builder).ParentID(thor.Bytes32{1}).Timestamp(1).GasLimit(100000).Beneficiary(thor.Address{1}).GasUsed(11234).TotalScore(1).StateRoot(thor.Bytes32{1}).ReceiptsRoot(thor.Bytes32{1}).Build()

header := block.Header()
```
obviously the builder logic can be responsibility of the moved function

paologalligit (changes requested):

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,64 @@
...
+	bandwidth := Bandwidth{
+		value: 0,
+		lock:  sync.Mutex{},
+	}
+
+	header := block.NewMockHeader()
+	bandwidth.Update(header, 1)
+	val := bandwidth.Value()
+
+	if val != 11234000000000 {
```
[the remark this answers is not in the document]
MakisChristou: Yes thats the expected value.

block/header.go
MakisChristou: I see I didn't know you can use the builder.

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,73 @@
...
+		value: 0,
+		lock:  sync.Mutex{},
+	}
+
+	block := new(block.Builder).ParentID(thor.Bytes32{1}).Timestamp(1).GasLimit(100000).Beneficiary(thor.Address{1}).GasUsed(11234).TotalScore(1).StateRoot(thor.Bytes32{1}).ReceiptsRoot(thor.Bytes32{1}).Build()
+	header := block.Header()
+	bandwidth.Update(header, 1)
+	val := bandwidth.SuggestGasLimit()
+
+	if val != 5617000000000 {
```
paologalligit, line 70: Here the message below should state "Expected 5617000000000..." and not 0. But since this is a test I would assert the value must equal the expected value, so for all of the three tests replace the `if` statements with `assert.Equal(t, expectedValue, actualValue)`

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) 2020 The VeChainThor developers
```
paologalligit, line 1: Since this is a new file should be `2024`

paologalligit (changes requested):
Last thing, remove also the whitespaces the linter is complaining about, so we're good to merge it

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,70 @@
...
+		lock:  sync.Mutex{},
+	}
+
+	block := new(block.Builder).ParentID(thor.Bytes32{1}).Timestamp(1).GasLimit(100000).Beneficiary(thor.Address{1}).GasUsed(11234).TotalScore(1).StateRoot(thor.Bytes32{1}).ReceiptsRoot(thor.Bytes32{1}).Build()
+	header := block.Header()
+	bandwidth.Update(header, 1)
+	val := bandwidth.SuggestGasLimit()
+
+	if val != 5617000000000 {
+		t.Errorf("Expected 0, got %d", val)
```
paologalligit, line 68: The message here is still wrong

paologalligit (dismissed):
Just fix the comment, then it's ready to be merged 👍

libotony (dismissed):

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,70 @@
...
+	bandwidth := Bandwidth{
+		value: 0,
+		lock:  sync.Mutex{},
+	}
+
+	val := bandwidth.Value()
+
+	if val != 0 {
+		t.Errorf("Expected 0, got %d", val)
+	}
```
libotony, line 27:
```suggestion

        assert.Equal(t, uint64(0), val)
```

libotony, line 27:
```suggestion
	assert.Equal(t, uint64(0), val)
```

cmd/thor/bandwidth/bandwidth_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	block := new(block.Builder).ParentID(thor.Bytes32{1}).Timestamp(1).GasLimit(100000).Beneficiary(thor.Address{1}).GasUsed(11234).TotalScore(1).StateRoot(thor.Bytes32{1}).ReceiptsRoot(thor.Bytes32{1}).Build()
+	header := block.Header()
+	bandwidth.Update(header, 1)
+	val := bandwidth.SuggestGasLimit()
+
+	if val != 5617000000000 {
+		t.Errorf("Expected 5617000000000, got %d", val)
+	}
```
libotony, line 69:
```suggestion
	assert.Equal(t, uint64(5617000000000), val)
```

libotony (approved):

PR #734, opened 2024-05-09
Adding api-logs-enabled flag

otherview:
# Description

Part of the Replay feature functionality - Ticket https://github.com/vechain/protocol-board-repo/issues/183
This flag adds the ability to log incoming requests.
Logs are also displayed if the log level is set to debug.
It defaults to NOT ENABLED.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

Added unit tests.

# Checklist:


cmd/thor/flags.go
```diff
@@ -59,6 +59,10 @@ var (
 		Name:  "api-allow-custom-tracer",
 		Usage: "allow custom JS tracer to be used tracer API",
 	}
+	apiLogsEnabledFlag = cli.BoolFlag{
```
darrenvechain, line 62: Should it be `cli.StringFlag` and the user specifies the log file location?

go.mod
```diff
@@ -18,11 +18,12 @@ require (
 	github.com/mattn/go-isatty v0.0.3
 	github.com/mattn/go-sqlite3 v1.14.9
 	github.com/mattn/go-tty v0.0.0-20180219170247-931426f7535a
+	github.com/otherview/filerotatewriter v0.0.0-20240509103702-a305c09fd53a
```
darrenvechain, line 21: Do we need a lib for this? Would personally prefer to have less dependencies

otherview: Yeah I know you mean, I didn't find a place that would make sense, so decided to use it as an external lib. We do need a package to wrap the rotation logic.
Where do you think it should be located ?

darrenvechain: Possibly `./api/logging` ?

darrenvechain: Actually thats not nice either, because then its under a `logging` package name.

cmd/thor/flags.go
otherview: I think we can probably add that flag if this feature takes traction. But I imagine it's easier to have this boolean flag with a known directory (`instaceDir`) instead of a three option flag (DefaultDir, Specify `null` to disable or Specify `path` to enable feature and set the log path). How does that sound to you ?

go.mod
otherview: Yeah I was resistant to add that there because it would "ruin" the nice packaging of the `api` folder. What about `api/utils/logfile` ?

darrenvechain:
@otherview does this persist all requests indefinitely ? Maybe we should clear logs older than X time or once they've reached storage of Y GB?

otherview:
> @otherview does this persist all requests indefinitely ? Maybe we should have a daily cleanup action that clears log files older than X time? Or once it's using more the Y GB of storage

It rotates files, a max of 10 files of 10mbs each.
I'll add this to the comments 👍

darrenvechain:
> > @otherview does this persist all requests indefinitely ? Maybe we should have a daily cleanup action that clears log files older than X time? Or once it's using more the Y GB of storage
> 
> It rotates files, a max of 10 files of 10mbs each. I'll add this to the comments 👍

Should we make it configurable? I feel like this would be used very quickly on @kgapos mainnet nodes

otherview:
> > > @otherview does this persist all requests indefinitely ? Maybe we should have a daily cleanup action that clears log files older than X time? Or once it's using more the Y GB of storage
> > 
> > 
> > It rotates files, a max of 10 files of 10mbs each. I'll add this to the comments 👍
> 
> Should we make it configurable with defaults? I feel like this would be used very quickly on @kgapos mainnet nodes

Yeap, done 👍

go.mod
darrenvechain: Makes sense to me

otherview: Done 👍

darrenvechain:
@otherview IMO we probably don't need to log requests to the console, as it would make debugging quite difficult in nodes with a lot a traffic

image

api/utils/rotatewriter/rotate_writer.go
```diff
@@ -0,0 +1,132 @@
...
+	// Rotate the log file if it exceeds the maximum file size
+	if w.currentSize+int64(len(p)) > w.maxFileSize {
+		err := w.rotateLogFile()
+		if err != nil {
+			return 0, err
+		}
+	}
+
+	// Write the log message to the current log file
+	n, err := w.currentFile.Write(p)
```
darrenvechain, line 56: Should we do?

```go
n, err := w.currentFile.Write(append(p, '\n'))
```

image

otherview: Damn, I thought I pushed that this morning.
We should defo append the \n, but if we do it in the api_request_logger, then the writer becomes agnostic of format.

libotony:
I would prefer the feature to be done outside of the protocol client. Such as reverse proxy with log enabled.

otherview:
Feature scope change, the api request logs are now issued to the log stdou following the same structure as other logs in the app.

api/api_request_logger.go
```diff
@@ -0,0 +1,121 @@
...
+type RequestLogger struct {
+	enabled      bool
+	writerChan   chan entry
+	stopChan     chan bool
+	outputWriter func(msg string, ctx ...interface{})
+}
+
+// NewThorRequestLogger applies thor business logic
+// when returning an instance of *RequestLogger
+func NewThorRequestLogger(enabledFlag bool, enableLog bool) *RequestLogger {
```
libotony, line 28: This looks wired to me

api/api.go
```diff
@@ -43,6 +43,7 @@ func New(
 	pprofOn bool,
 	skipLogs bool,
 	allowCustomTracer bool,
+	reqLogger *RequestLogger,
```
libotony, line 46: Can this be a bool?

api/api_request_logger.go
```diff
@@ -0,0 +1,121 @@
...
+
+// Handle returns an http handler to ensure requests are syphoned into the writer
+func (l *RequestLogger) Handle(handler http.Handler) http.Handler {
+	fn := func(w http.ResponseWriter, r *http.Request) {
+		// Read and log the body (note: this can only be done once)
+		// Ensure you don't disrupt the request body for handlers that need to read it
+		var bodyBytes []byte
+		var err error
+		if r.Body != nil {
+			bodyBytes, err = io.ReadAll(r.Body)
```
libotony, line 59: Close the body

api/api_request_logger.go
```diff
@@ -0,0 +1,121 @@
...
+	// make sure any pending message is written
+	l.stopChan <- true
+	<-l.stopChan
+}
+
+func (l *RequestLogger) start() {
+	go func() {
+		for {
+			select {
+			case item := <-l.writerChan:
```
libotony, line 100: Instead of building a queue inside the request logger, how about delegating the queue to the logger function, I found that logger handler can be configured to a buffered handler.

api/api_request_logger.go
```diff
@@ -0,0 +1,121 @@
...
+			case <-l.stopChan:
+				close(l.stopChan)
+				return
+			}
+		}
+	}()
+}
+
+type entry struct {
+	Timestamp time.Time `json:"timestamp"`
```
libotony, line 117: The timestamp is there in the log, seems redundant
```
INFO[05-17|15:23:46] [request] {"timestamp":"2024-05-17T15:23:46.036722+08:00","uri":"/accounts/*","method":"POST","body":"{\n  \"gas\": 50000,\n  \"gasPrice\": \"1000000000000000\",\n  \"caller\": \"0x6d95e6dca01d109882fe1726a2fb9865fa41e7aa\",\n  \"provedWork\": \"1000\",\n  \"gasPayer\": \"0xd3ae78222beadb038203be21ed5ce7c9b1bff602\",\n  \"expiration\": 1000,\n  \"blockRef\": \"0x00000000851caf3c\",\n  \"clauses\": [\n    {\n      \"to\": \"0x0000000000000000000000000000456E65726779\",\n      \"value\": \"0x0\",\n      \"data\": \"0xa9059cbb0000000000000000000000000f872421dc479f3c11edd89512731814d0598db50000000000000000000000000000000000000000000000013f306a2409fc0000\"\n    },\n    {\n      \"to\": \"0xf077b491b355E64048cE21E3A6Fc4751eEeA77fa\",\n      \"value\": \"0x6124fee993bc00000\",\n      \"data\": \"0x\"\n    },\n    {\n      \"to\": null,\n      \"value\": \"0x0\",\n      \"data\": \"0x6080604052348015600f57600080fd5b50609f8061001e6000396000f300608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff1680631820cabb146044575b600080fd5b348015604f57600080fd5b506056606c565b6040518082815260200191505060405180910390f35b62015180815600a165627a7a723058200ac7475da248e2fc26c057319e296e90c24d5f8b9bf956fb3b77545642cad3b10029\"\n    }\n  ]\n}"} ```

File: api/api_request_logger.go [hunk 9]
```diff
@@ -0,0 +1,121 @@
...
+		for {
+			select {
+			case item := <-l.writerChan:
+				marshal, err := json.Marshal(item)
+				if err != nil {
+					log.Warn("unable to marshal api request entry", "err", err)
+					continue
+				}
+
+				l.outputWriter("[request] " + string(marshal))
```
libotony, line 107: `log  = log15.New("pkg", "api")` might be good for filtering and can this be done via the traditional formatted way as below? 
```
log.Debug("URI", item.URI, "Method", item.Method)
```

api/api_request_logger.go
otherview: Understood, however the timestamp of the log would need to be more accurate, and there might be differences between the event time and the time when it was logged.

api/api_request_logger.go
otherview: Yes, that was my first go-to, but I was looking for something prepended vs appended so that it's a fixed structure text instead of a variable one.

Ex. `INFO[05-17|15:23:46] [request] {"timestamp":` its easy to filter on `[request]`
Ex. `INFO[05-17|15:23:46] {"timestamp":.......WEIRD THINGS} type=request` it's not easy to filter on `[request]`

otherview: Perhaps if it's easy to prepend the tags? It seem'd convoluted to me.

api/api_request_logger.go
otherview: Perhaps this might look weird because this type of business logic used to be existing in files like the `utils.go`. 

Arguably, files like `utils.go` should not hold the thor business logic on how to instantiate a package.
It make it hard to test the logic and the most important instantiation logic is pushed into a gather-all kind of file.
ex. The [flags around the p2p](https://github.com/vechain/thor/blob/master/cmd/thor/utils.go#L446-L496)

Typically, the way that I've seen this done is by having the package implement a `NewThorBusinessLogicX` that holds the business logic. That way it's possible to create unit tests for the package AND for the business logic.

api/api.go
otherview: It would have to be 2 bools ?

api/api_request_logger.go
otherview: I don't mind, but would like to avoid marshalling overhead on the request proxy, would that still be possible ?

api/api_request_logger.go
otherview: Turns out https://pkg.go.dev/net/http#Request
```
        // For server requests, the Request Body is always non-nil
	// but will return EOF immediately when no body is present.
	// The Server will close the request body. The ServeHTTP
	// Handler does not need to.
```

libotony: Since you have overwritten the body, possibly memory leak will happen if io.ReadAll fails and not closed

api/api_request_logger.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2024 The VeChainThor developers
```
libotony, line 1: File name?

cmd/thor/flags.go
```diff
@@ -59,6 +59,10 @@ var (
...
+		Name:  "api-logs-enabled",
+		Usage: "logs inbound requests to stdout",
```
libotony, line 64:
```suggestion
		Usage: "logs API requests to stdout",
```

api/api.go
```diff
@@ -92,11 +94,15 @@ func New(
 	}
 
 	handler := handlers.CompressHandler(router)
+	if isReqLoggerEnabled {
+		handler = RequestLoggerHandler(handler, log15.New("pkg", "api"))
```
libotony, line 98: `log15.New("pkg", "api")` should be a package level global variable

otherview: yeah fair point.

api/api_request_logger.go
otherview: removed the `api_` 👍

libotony (approved):

api/api_request_logger.go
otherview: It should return before the r.Body is written to.

libotony (approved):

PR #740, opened 2024-05-14
Selectable block interval on solo

MakisChristou:
Tackles the following [issue](https://github.com/orgs/vechain/projects/22/views/1?filterQuery=label%3A%22App%3A+Thor%22%2C%22App%3A+VIP%22%2C%22App%3A+Thorest%22++solo&pane=issue&itemId=53228366). In this PR I basically added the option to use a command-line argument specifically for solo to allow the developer to choose a custom block interval with a minimum value of 1. I tested it and it looks like its working fine for 1 second block intervals as well as larger intervals like 100 seconds.

### DIFF
--- a/cmd/thor/main.go
+++ b/cmd/thor/main.go
@@ -335,8 +335,7 @@ func soloAction(ctx *cli.Context) error {
 
 	blockInterval := ctx.Int(blockInterval.Name)
 	if blockInterval == 0 {
-		blockInterval = 10
-		return fmt.Errorf("block-interval cannot be zero")
+		return errors.New("block-interval cannot be zero")
 	}
 
 	printSoloStartupMessage(gene, repo, instanceDir, apiURL, forkConfig)
--- a/cmd/thor/main.go
+++ b/cmd/thor/main.go
@@ -105,6 +105,7 @@ func main() {
 					apiBacktraceLimitFlag,
 					apiAllowCustomTracerFlag,
 					onDemandFlag,
+					blockInterval,
 					persistFlag,
 					gasLimitFlag,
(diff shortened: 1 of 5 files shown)


claytonneal:
Can we add to the readme please?

cmd/thor/main.go
```diff
@@ -332,6 +333,12 @@ func soloAction(ctx *cli.Context) error {
 	}
 	defer func() { log.Info("stopping API server..."); srvCloser() }()
 
+	blockInterval := ctx.Int(blockInterval.Name)
+	if blockInterval == 0 {
+		blockInterval = 10
+		return fmt.Errorf("block-interval cannot be zero")
```
libotony, line 339:
```suggestion
		return errors.New("block-interval cannot be zero")
```

cmd/thor/solo/solo_test.go
```diff
@@ -29,7 +29,7 @@ func newSolo() *Solo {
 	repo, _ := chain.NewRepository(db, b)
 	mempool := txpool.New(repo, stater, txpool.Options{Limit: 10000, LimitPerAccount: 16, MaxLifetime: 10 * time.Minute})
 
-	return New(repo, stater, logDb, mempool, 0, true, false, thor.ForkConfig{})
+	return New(repo, stater, logDb, mempool, 0, true, false, 10, thor.ForkConfig{})
```
libotony, line 32:
```suggestion
	return New(repo, stater, logDb, mempool, 0, true, false, thor.BlockInterval, thor.ForkConfig{})
```

libotony (approved):

PR #742, opened 2024-05-17
remove flaky subscription test

paologalligit:
# Description

We've had multiple failing tests in ci/cd caused by a timeout while running the subscriptions tests. The issue is not directly reproducible locally and failing runs happen randomly, but the issue seems to be related to the websocket connection not being able to receive a message, so it hangs waiting indefinitely. This PR adds a timeout of 5 seconds: if a message is not received the test is simply skipped.

Fixes [issue-199](https://github.com/vechain/protocol-board-repo/issues/199)

## Type of change

# Checklist:

### DIFF
--- a/api/subscriptions/subscriptions_test.go
+++ b/api/subscriptions/subscriptions_test.go
@@ -77,8 +77,8 @@ func testHandlePendingTransactions(t *testing.T) {
 		} else {
 			assert.Equal(t, transaction.ID(), pendingTx.ID)
 		}
-	case <-time.After(5 * time.Second):
-		// If in 5 seconds the message is not received, the test will be skipped
+	case <-time.After(time.Second):
+		// If in 1 second the message is not received, the test will be skipped
 		t.SkipNow()
 	}
 }
--- a/api/subscriptions/subscriptions_test.go
+++ b/api/subscriptions/subscriptions_test.go
@@ -13,6 +13,7 @@ import (
 	"net/url"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/gorilla/mux"
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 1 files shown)


api/subscriptions/subscriptions_test.go
```diff
@@ -59,15 +60,26 @@ func testHandlePendingTransactions(t *testing.T) {
...
+	select {
+	case <-readComplete:
+		assert.NoError(t, err)
+		var pendingTx *PendingTxIDMessage
+		if err := json.Unmarshal(msg, &pendingTx); err != nil {
+			t.Fatal(err)
+		} else {
+			assert.Equal(t, transaction.ID(), pendingTx.ID)
+		}
+	case <-time.After(5 * time.Second):
```
libotony, line 80: The test case runs fast, how about making it 1 second?

api/subscriptions/subscriptions_test.go
```diff
@@ -59,15 +60,26 @@ func testHandlePendingTransactions(t *testing.T) {
...
+		assert.NoError(t, err)
+		var pendingTx *PendingTxIDMessage
+		if err := json.Unmarshal(msg, &pendingTx); err != nil {
+			t.Fatal(err)
+		} else {
+			assert.Equal(t, transaction.ID(), pendingTx.ID)
+		}
+	case <-time.After(time.Second):
+		// If in 1 second the message is not received, the test will be skipped
+		t.SkipNow()
```
leszek-vechain, line 82: should we rather fail it ?

paologalligit: I would not make it fail since, when the time's up, it's not a sut failure, but the premises were missing. I noticed that the timeout happens when the tx is not added into the mempool and the websocket hangs indefinitely. Another solution would be to assert the mempool is not empty, wait until we have the tx, and then read with the websocket. Let me know if you think this is a better approach.

libotony (approved):

PR #744, opened 2024-05-20
tweak metrics in txpool and node

libotony:
# Description

# Checklist:

### DIFF
--- a/bft/engine.go
+++ b/bft/engine.go
@@ -132,7 +132,7 @@ func (engine *BFTEngine) CommitBlock(header *block.Header, isPacking bool) error
 				return err
 			}
 			engine.finalized.Store(id)
-			metricBlocksCommitted().AddWithLabel(1, map[string]string{"status": "finalized"})
+			metricBlocksCommitted().Add(1)
 		}
 	}
 
--- a/bft/engine.go
+++ b/bft/engine.go
@@ -27,7 +27,7 @@ const dataStoreName = "bft.engine"
 var (
 	finalizedKey = []byte("finalized")
 
-	metricBlocksCommitted = metrics.LazyLoadCounterVec("block_bft_committed_count", []string{"status"})
+	metricBlocksCommitted = metrics.LazyLoadCounterVec("bft_committed_count", []string{"status"})
 )
 
 type Finalizer interface {
@@ -153,7 +153,6 @@ func (engine *BFTEngine) CommitBlock(header *block.Header, isPacking bool) error
 			return err
 		}
 		engine.casts.Mark(checkpoint, state.Quality)
-		metricBlocksCommitted().AddWithLabel(1, map[string]string{"status": "proposed"})
 	}
 
 	return nil
--- a/bft/engine.go
+++ b/bft/engine.go
@@ -14,7 +14,6 @@ import (
 	"github.com/vechain/thor/v2/cache"
 	"github.com/vechain/thor/v2/chain"
 	"github.com/vechain/thor/v2/kv"
-	"github.com/vechain/thor/v2/metrics"
 	"github.com/vechain/thor/v2/muxdb"
 	"github.com/vechain/thor/v2/state"
 	"github.com/vechain/thor/v2/thor"
@@ -24,11 +23,7 @@ import (
 
 const dataStoreName = "bft.engine"
 
-var (
-	finalizedKey = []byte("finalized")
-
-	metricBlocksCommitted = metrics.LazyLoadCounterVec("bft_committed_count", []string{"status"})
-)
+var finalizedKey = []byte("finalized")
 
 type Finalizer interface {
 	Finalized() thor.Bytes32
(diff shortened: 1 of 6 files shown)


cmd/thor/node/packer_loop.go
```diff
@@ -204,6 +207,10 @@ func (n *Node) pack(flow *packer.Flow) error {
 		if v, updated := n.bandwidth.Update(newBlock.Header(), time.Duration(realElapsed)); updated {
 			log.Debug("bandwidth updated", "gps", v)
 		}
+
+		metricBlockProcessedTxs().Add(int64(len(receipts)))
+		metricBlockProcessedGas().Add(int64(newBlock.Header().GasUsed()))
+		metricBlockProcessedDuration().Observe(time.Duration(realElapsed).Milliseconds())
```
otherview, line 213: Does this distinguish between tx metrics from packed blocks and tx metrics from received blocks ?

txpool/tx_pool.go
```diff
@@ -132,6 +132,7 @@ func (p *TxPool) housekeeping() {
 					p.executables.Store(executables)
 				}
 
+				metricTxPoolGauge().GaugeWithLabel(0-int64(removed), map[string]string{"source": "washed", "total": "true"})
 				log.Debug("wash done", ctx...)
```
otherview, line 136: Should work no ?
```suggestion
				metricTxPoolGauge().GaugeWithLabel(-int64(removed), map[string]string{"source": "washed", "total": "true"})
				log.Debug("wash done", ctx...)
```

bft/engine.go
```diff
@@ -153,7 +153,6 @@ func (engine *BFTEngine) CommitBlock(header *block.Header, isPacking bool) error
 			return err
 		}
 		engine.casts.Mark(checkpoint, state.Quality)
-		metricBlocksCommitted().AddWithLabel(1, map[string]string{"status": "proposed"})
```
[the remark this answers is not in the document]
libotony: block proposed already captured in node metrics, `metricBlockProcessedCount` with label `type=proposed`

cmd/thor/node/packer_loop.go
libotony: No, I would consider the proposed block count as necessary but tx and gas are not.

bft/engine.go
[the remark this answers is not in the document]
otherview: Are processed and committed the same ?

cmd/thor/node/packer_loop.go
otherview: Why not? Would it not be interesting to see the tx count and gas sum of proposed blocks ?

bft/engine.go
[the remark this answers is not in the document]
libotony: > Are processed and committed the same ?

Not the same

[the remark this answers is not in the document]
otherview: I don't think this works the way I thought it worked.. That's cool.

So I guess there's no point in having `status=finalized` if it's always the same status ?

cmd/thor/node/packer_loop.go
```diff
@@ -106,16 +106,21 @@ func (n *Node) packerLoop(ctx context.Context) {
...
+				n.txPool.Remove(tx.Hash(), tx.ID())
+			}
+			metricBlockProcessedCount().AddWithLabel(1, map[string]string{"type": "proposed", "success": "true"})
+		} else {
+			metricBlockProcessedCount().AddWithLabel(1, map[string]string{"type": "proposed", "success": "false"})
 		}
 	}()
 
-	return evalBlockProposeMetrics(n.guardBlockProcessing(flow.Number(), func(conflicts uint32) error {
+	return n.guardBlockProcessing(flow.Number(), func(conflicts uint32) error {
```
otherview, line 123: I like it, it's better than what it was, however it's not a simple flow to understand.
Just a thought, but perhaps it wouldn't be a bad idea for us to consider refactoring this in the future, worth putting a todo here?

otherview (approved):
lgtm - before merging would consider removing the status from the `bft_committed_count` and add a few comments or todos

cmd/thor/node/packer_loop.go
libotony: How about capturing it in a ticket? Seems fine to me.

otherview: Yep, created https://github.com/vechain/protocol-board-repo/issues/209 👍

cmd/thor/node/node.go
```diff
@@ -402,11 +402,14 @@ func (n *Node) processBlock(newBlock *block.Block, stats *blockStats) (bool, err
...
 		switch {
-		case err == errKnownBlock || err == errBFTRejected:
+		case err == errKnownBlock:
 			stats.UpdateIgnored(1)
 			return false, nil
 		case consensus.IsFutureBlock(err) || err == errParentMissing || err == errBlockTemporaryUnprocessable:
 			stats.UpdateQueued(1)
+		case err == errBFTRejected:
+			// TODO: capture metrics
+			log.Debug(fmt.Sprintf("block rejected by BFT engine\n%v", newBlock.Header()))
```
otherview, line 412: Nice 👍

PR #777, opened 2024-06-20
add api-logs-limit flag

libotony:
# Description

Add a flag to allow node runner to specify the logs limit for the logs filter API.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:

### DIFF
--- a/cmd/thor/flags.go
+++ b/cmd/thor/flags.go
@@ -59,6 +59,11 @@ var (
 		Name:  "api-allow-custom-tracer",
 		Usage: "allow custom JS tracer to be used tracer API",
 	}
+	apiLogsLimitFlag = cli.IntFlag{
+		Name:  "api-logs-limit",
+		Value: 1000,
+		Usage: "limit the number of logs returned by /logs API",
+	}
 	enableAPILogsFlag = cli.BoolFlag{
 		Name:  "enable-api-logs",
 		Usage: "enables API requests logging",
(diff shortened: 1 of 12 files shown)


darrenvechain:
Should we return an API error if the user specify more than the limit? Because if the user requests 5000 and gets 1000, then it looks like there are no more results, and they will stop making paginated requests

cmd/thor/flags.go
```diff
@@ -59,6 +59,11 @@ var (
 		Name:  "api-allow-custom-tracer",
 		Usage: "allow custom JS tracer to be used tracer API",
 	}
+	apiLogsLimitFlag = cli.IntFlag{
```
leszek-vechain, line 62: should we add it too docs, or it is hidden ?

libotony: Will add to the docs, thanks!

otherview (commented):
Structure is solid, and the limit makes sense to happen at the api lvl.

Probably worth to have a unit test that ensures the filtering is returning the max number of results in the `events_test.go` and in the `transfers_test.go`. We're testing the `NormalizeOptions` but if for example that function gets left out for some reason the tests won't capture that the query becomes now unbounded ?

Also, perhaps update the api yml definition to account for the existence of the flag ?

darrenvechain:
@libotony I made a PR here to fix the E2E. We need to merge this to main and then put the commit SHA in the e2e GHA

https://github.com/vechain/thor-e2e-tests/pull/24/files

otherview (approved):

PR #779, opened 2024-06-26
use uint64 for number flags

libotony:
# Description

This is a minor improvement of the flags parsing component.

# Checklist:

### DIFF
--- a/cmd/thor/utils.go
+++ b/cmd/thor/utils.go
@@ -60,9 +60,8 @@ import (
 
 var devNetGenesisID = genesis.NewDevnet().ID()
 
-func initLogger(ctx *cli.Context) {
-	logLevel := ctx.Int(verbosityFlag.Name)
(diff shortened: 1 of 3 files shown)


cmd/thor/utils.go
```diff
@@ -712,3 +711,14 @@ func parseNodeList(list string) ([]*discover.Node, error) {
 
 	return nodes, nil
 }
+
+func readIntFromFlag(ctx *cli.Context, name string) (int, error) {
```
otherview, line 715: I think it's a good idea to normalize this, perhaps the behaviour can be simplified further by passing the uint64 ?

```suggestion
func readIntFromUInt64Flag(val uint64) (int, error) {
```

libotony: The name of the flag was used for a better error hint for that, that's why it comes with the context, naming seems better, but do you consider passing `(val uint64, name string)` is better?

otherview (approved):

PR #799, opened 2024-07-25
Ability to change log verbosity while the node is running

MakisChristou:


cmd/thor/utils.go
```diff
@@ -60,21 +61,25 @@ import (
...
+	var level slog.LevelVar
+	level.Set(logLevel)
 
 	var handler slog.Handler
 	if jsonLogs {
 		handler = log.JSONHandlerWithLevel(output, logLevel)
 	} else {
 		useColor := (isatty.IsTerminal(os.Stderr.Fd()) || isatty.IsCygwinTerminal(os.Stderr.Fd())) && os.Getenv("TERM") != "dumb"
-		handler = log.NewTerminalHandlerWithLevel(output, logLevel, useColor)
+		handler = log.NewTerminalHandlerWithLevel(output, &level, useColor)
```
darrenvechain, line 75: @MakisChristou can you also update the JSON handler ?

admin/admin.go
```diff
@@ -0,0 +1,43 @@
...
+	"net/http"
+
+	"github.com/gorilla/handlers"
+	"github.com/gorilla/mux"
+	"github.com/vechain/thor/v2/log"
+)
+
+func HTTPHandler(logLevel *slog.LevelVar) http.Handler {
+	router := mux.NewRouter()
+	router.PathPrefix("/admin").Handler(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
darrenvechain, line 20: @MakisChristou IMO this should be a post request and the level should be declared in the request body.

GET requests shouldn't be used to update state


Maybe:
- GET `/admin/loglevel` to get the current log level
- POST `/admin/loglevel` with a body of `{ "level": "info" `

This allow us to create many endpoints on the admin domain if we have a use case to do so in the future

cmd/thor/main.go
```diff
@@ -170,6 +174,16 @@ func defaultAction(ctx *cli.Context) error {
 		defer func() { log.Info("stopping metrics server..."); close() }()
 	}
 
+	adminURL := ""
+	if ctx.Bool(enableAdminFlag.Name) {
+		url, close, err := startAdminServer(ctx.String(adminAddrFlag.Name), logLevel)
```
darrenvechain, line 179: Do we have a default address in case one is not provided? (Same pattern as metrics)

cmd/thor/flags.go
```diff
@@ -150,6 +150,15 @@ var (
 		Value: "localhost:2112",
 		Usage: "metrics service listening address",
 	}
+	enableAdminFlag = cli.BoolFlag{
```
leszek-vechain, line 153: should we put it into documentation or these flags are hidden ?

cmd/thor/main.go
leszek-vechain: additionally can we check if it does not hijack websockets :D (not sure where to put it)

MakisChristou: @darrenvechain The default address is `http://localhost:2113` which is defined in the cli arguments if I am not mistaken

MakisChristou: @leszek-vechain What do you mean by "hijack websockets"? 🤔

cmd/thor/flags.go
MakisChristou: Sure I can add it

cmd/thor/main.go
leszek-vechain: when Pedro did add the server for metrics I think, websockets were redirected there if I correctly remember, so just trying to avoid it happening again

cmd/thor/flags.go
MakisChristou: Btw when looking into the docs I saw this:

Metrics is enabled in nodes by default. It's possible to disable it by setting  `--enable-metrics=false`.

Isn't this invalid as of now?

admin/admin.go
MakisChristou: Added this change

cmd/thor/main.go
MakisChristou: I am not familiar with that happened there. I just created a new server on a different port. Hopefully it should not affect anything else.

cmd/thor/flags.go
darrenvechain: Ya I think this is wrong

MakisChristou: Should it be fixed in this PR?

docs/hosting-a-node.md
```diff
@@ -101,11 +101,29 @@ _As of 22nd April 2024, a full node without logs uses **~100 GB** of disk space.
 Telemetry plays a critical role in monitoring and managing blockchain nodes efficiently. 
 Below is an overview of how metrics is integrated and utilized within our node systems.
 
-Metrics is enabled in nodes by default. It's possible to disable it by setting  `--enable-metrics=false`.
-By default, a [prometheus](https://prometheus.io/docs/introduction/overview/) server is available at `localhost:2112/metrics` with the metrics.
+Metrics is not enabled in nodes by default. It's possible to enable it by setting  `--enable-metrics`.
```
leszek-vechain, line 104: just wonder if it should be is or are

MakisChristou: I would guess its the metrics server so thats why it was is?

leszek-vechain: probably question to @nwbrettski or someone else with better english understanding than mine 😆 just it was not feeling right for me, thats why wanted to double check

nwbrettski: Suggest modifying it to be

Metrics are not enabled in nodes by default. You can enable them by setting --enable-metrics.

admin/admin.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func getLogLevelHandler(logLevel *slog.LevelVar) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		response := logLevelResponse{
+			CurrentLevel: logLevel.Level().String(),
+		}
+		w.Header().Set("Content-Type", "application/json")
+		if err := json.NewEncoder(w).Encode(response); err != nil {
+			http.Error(w, "Failed to encode response", http.StatusInternalServerError)
```
otherview, line 34: I think it would be nice to have an always json-format responses.
For payloads, responses and error responses everything as a json, seems to me, that would make easier than a json sometimes and a plain text on other times ?

cmd/thor/flags.go
otherview: No need to change. AFAIK the decision was made to leave metrics enabled per default given that the default server was bounded to localhost on a different port.

cmd/disco/utils.go
```diff
@@ -19,15 +20,19 @@ import (
 	"github.com/vechain/thor/v2/log"
 )
 
-func initLogger(lvl int) {
+func initLogger(lvl int) *slog.LevelVar {
 	logLevel := log.FromLegacyLevel(lvl)
+	var level slog.LevelVar
+	level.Set(logLevel)
```
otherview, line 26: Out of curiosity, do we want the admin endpoint in disco ? I wonder if that would make sense.

docs/hosting-a-node.md
```diff
@@ -101,11 +101,29 @@ _As of 22nd April 2024, a full node without logs uses **~100 GB** of disk space.
 Telemetry plays a critical role in monitoring and managing blockchain nodes efficiently. 
 Below is an overview of how metrics is integrated and utilized within our node systems.
 
-Metrics is enabled in nodes by default. It's possible to disable it by setting  `--enable-metrics=false`.
-By default, a [prometheus](https://prometheus.io/docs/introduction/overview/) server is available at `localhost:2112/metrics` with the metrics.
+Metrics are not enabled in nodes by default. It's possible to enable them by setting  `--enable-metrics`.
+Once enabled, a [prometheus](https://prometheus.io/docs/introduction/overview/) server is available at `localhost:2112/metrics` with the metrics.
```
otherview, line 105: Let's avoid changing this in this PR

otherview (commented):
Looks good - ooc is there a way that we can test this feature on unit tests ?

cmd/thor/flags.go
MakisChristou: So no documentation for the new flags? @otherview

cmd/disco/utils.go
MakisChristou: Not sure about the answer to this. @libotony , @darrenvechain Do we need to change the log verbosity of disco at runtime?

darrenvechain: IMO it's not needed, but I don't feel too strongly either way

docs/hosting-a-node.md
```diff
@@ -108,4 +108,22 @@ By default, a [prometheus](https://prometheus.io/docs/introduction/overview/) se
 curl localhost:2112/metrics
 ```
 
-Instrumentation is in a beta phase at this stage. You can read more about the metric types [here](https://prometheus.io/docs/concepts/metric_types/).
\ No newline at end of file
+Instrumentation is in a beta phase at this stage. You can read more about the metric types [here](https://prometheus.io/docs/concepts/metric_types/).
+
+### Admin
+
+Admin is used to allow privilidged actions to the node by the administrator. Currently is supports changing the logger's verbosity at runtime.
```
darrenvechain, line 115:
```suggestion
Admin is used to allow privileged actions to the node by the administrator. Currently it supports changing the logger's verbosity at runtime.
```

otherview (commented):
Overall lgtm, I am starting to think that it can be benefitial to follow the same API struct pattern as the api package and perhaps having a couple of kickstarter unit tests ?

cmd/thor/utils.go
```diff
@@ -587,6 +592,27 @@ func startMetricsServer(addr string) (string, func(), error) {
 	}, nil
 }
 
+func startAdminServer(addr string, logLevel *slog.LevelVar) (string, func(), error) {
```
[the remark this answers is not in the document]
MakisChristou: Not sure. For this I just used the same pattern that is used in metrics with `startMetricsServer` living in `utils.go`.

admin/admin.go
```diff
@@ -0,0 +1,103 @@
...
+			writeError(w, http.StatusMethodNotAllowed, "method not allowed")
+		}
+	}
+}
+
+func HTTPHandler(logLevel *slog.LevelVar) http.Handler {
+	router := mux.NewRouter()
+	router.HandleFunc("/admin/loglevel", logLevelHandler(logLevel))
+	return handlers.CompressHandler(router)
+}
```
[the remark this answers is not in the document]
MakisChristou: Wouldn't it make more sense to use the same pattern as the metrics server since our goal is something similar? In the metric's case the `startMetricsServer` is in the `utils.go` file while the handler is returned from the `telemetry.go` file similar to my `admin.go` file. If not are you suggesting I move `startMetricsServer` into its own file basically?

otherview (dismissed):

leszek-vechain (dismissed):

leszek-vechain (approved):

PR #815, opened 2024-08-13
Add cli option for limit-per-account in the mempool

MakisChristou:
This PR makes the already existing flags 

- `txpool-limit-per-account` (default 16)
- `txpool-limit` (default 10k)

which were previously solo only flags available in testnet and mainnet modes. No unbounded option was added but we can discuss if that is necessary.

### DIFF
--- a/cmd/thor/flags.go
+++ b/cmd/thor/flags.go
@@ -150,11 +150,6 @@ var (
 		Value: "localhost:2112",
 		Usage: "metrics service listening address",
 	}
-	txPoolLimitFlag = cli.Uint64Flag{
-		Name:  "txpool-limit",
-		Value: 10000,
-		Usage: "set tx limit in pool",
-	}
 	txPoolLimitPerAccountFlag = cli.Uint64Flag{
 		Name:  "txpool-limit-per-account",
 		Value: 16,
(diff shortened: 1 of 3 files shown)


cmd/thor/flags.go
```diff
@@ -150,6 +150,16 @@ var (
 		Value: "localhost:2112",
 		Usage: "metrics service listening address",
 	}
+	txPoolLimitFlag = cli.Uint64Flag{
```
leszek-vechain, line 153: do we need to update docs as it not only solo flag now

cmd/thor/main.go
```diff
@@ -93,6 +93,8 @@ func main() {
 			disablePrunerFlag,
 			enableMetricsFlag,
 			metricsAddrFlag,
+			txPoolLimitFlag,
```
libotony, line 96: Txpool limit should be a solo-only flag.

otherview (approved):
lgtm 👍 

Do you think it is worth creating a test for the limit per account ?

leszek-vechain:
would be nice to have for sure, just to ensure that this change works as expected

MakisChristou:
> lgtm 👍
> 
> Do you think it is worth creating a test for the limit per account ?

Do you mean e2e tests or unit tests. Because there are already some unit tests for the account limit.

otherview:
> > lgtm 👍
> > Do you think it is worth creating a test for the limit per account ?
> 
> Do you mean e2e tests or unit tests. Because there are already some unit tests for the account limit.

Yeah, It guess that's fine, not worth the whole setup for the tx per account test. 
Good stuff !

PR #818, opened 2024-08-13
Thor client

paologalligit:
# Description

This PR introduces a thor client. This refers to this [gh issue](https://github.com/vechain/protocol-board-repo/issues/215).
The structure was taken from the [Draupnir's client](https://github.com/vechain/draupnir/tree/main/client).
The client supports the following operations:

- `GetTransactionReceipt`
- `InspectClauses`
- `InspectTxClauses`
- `SendTransaction`
- `SendEncodedTransaction`
- `GetLogEvents`
- `GetLogTransfer`
- `GetAccount`
- `GetAccountForRevision`
- `GetAccountCode`
- `GetAccountCodeForRevision`
- `GetStorage`
- `GetExpandedBlock`
- `GetBlock`
- `GetBestBlock`
- `GetTransaction`
- `GetPendingTransaction`
- `GetPeers`
- `ChainTag`
- `SubscribeBlocks`
- `SubscribeEvents`
- `SubscribeTransfers`
- `SubscribeTxPool`
- `SubscribeBeats`
- `SubscribeBeats2`

## Type of change

# How Has This Been Tested?

- [x] Unit tests

# Checklist:


thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package http
```
darrenvechain, line 6: http is a pretty commonly used package name, `client` or something else would be better IMO

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+	return &block, nil
+}
+
+func (c *Client) GetTransaction(txID *thor.Bytes32, isPending bool) (*transactions.Transaction, error) {
+	url := c.url + "/transactions/" + txID.String()
+	if isPending {
+		url += "?pending=true"
+	}
+
+	body, err := c.httpGET(url)
```
darrenvechain, line 180: `body` can be null here

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+	return &block, nil
+}
+
+func (c *Client) GetBlock(blockID string) (*blocks.JSONBlockSummary, error) {
+	body, err := c.httpGET(c.url + "/blocks/" + blockID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve block - %w", err)
+	}
+
+	var block blocks.JSONBlockSummary
```
darrenvechain, line 166: `body` can be null here, so maybe return not found error

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+	return c.httpGET(c.url + "/accounts/" + addr.String() + "/key/" + key.String())
+}
+
+func (c *Client) GetExpandedBlock(blockID string) (*blocks.JSONExpandedBlock, error) {
+	body, err := c.httpGET(c.url + "/blocks/" + blockID + "?expanded=true")
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve block - %w", err)
+	}
+
+	var block blocks.JSONExpandedBlock
```
darrenvechain, line 152: `body` can be null here, so maybe return not found error

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+func NewClient(url string) *Client {
+	return &Client{
+		url: url,
+		c:   &http.Client{},
+	}
+}
+
+func (c *Client) GetTransactionReceipt(txID *thor.Bytes32) (*transactions.Receipt, error) {
```
darrenvechain, line 36: Nothing wrong with the name, but I would have a preference to remove the `Get`s from the function names. Seems we have a lot of functions here prefixed with `Get`

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+
+	var txID common.TxSendResult
+	if err = json.Unmarshal(body, &txID); err != nil {
+		return nil, fmt.Errorf("unable to unmarshall inspection - %w", err)
+	}
+
+	return &txID, nil
+}
+
+func (c *Client) GetLogs(eventEndpoint string, req map[string]interface{}) ([]events.FilteredEvent, error) {
```
darrenvechain, line 82: Maybe `GetEvents` so it matches API name?

Also, `Get` gives me an ick since it is a `POST` request  :stuck_out_tongue_closed_eyes:

Maybe `FilterEvents`?

darrenvechain, line 82: Could we also have a typed req struct?

Maps are generally not too nice to work with

thorclient/http/client.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	return filteredEvents, nil
+}
+
+func (c *Client) GetLogsEvent(req map[string]interface{}) ([]events.FilteredEvent, error) {
+	return c.GetLogs(c.url+"/logs/event", req)
+}
+
+func (c *Client) GetAccount(addr *thor.Address, revision *thor.Bytes32) (*accounts.Account, error) {
```
darrenvechain, line 114: `revision` may not be a Bytes32, could be `best`, `finalized` or a number

thorclient/http/client.go
darrenvechain: Or maybe just move it inline with `thorclient`?

thorclient/http/client.go
paologalligit: Ahaha, that's something I didn't notice

libotony:
Can we abandon the API we have already marked as deprecated? Such as `SubscribeBeats `.

thorclient/thorclient.go
```diff
@@ -0,0 +1,212 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package client
```
darrenvechain, line 6: Package name does not match directory

thorclient/thorclient.go
```diff
@@ -0,0 +1,212 @@
...
+	"github.com/vechain/thor/v2/thorclient/wsclient"
+	"github.com/vechain/thor/v2/tx"
+)
+
+type Client struct {
+	conn   *httpclient.Client
+	wsConn *wsclient.Client
+}
+
+func NewClient(url string) *Client {
```
darrenvechain, line 33:
```suggestion
func New(url string) *Client {
```

darrenvechain (changes requested):

thorclient/thorclient.go
otherview: Fixed 👍

thorclient/wsclient/client.go
```diff
@@ -0,0 +1,143 @@
...
+	}, nil
+}
+
+func (c *Client) SubscribeEvents(query string) (<-chan *subscriptions.EventMessage, error) {
+	conn, err := c.connect("/subscriptions/event", query)
+	if err != nil {
+		return nil, fmt.Errorf("unable to connect - %w", err)
+	}
+
+	return subscribe[subscriptions.EventMessage](conn)
```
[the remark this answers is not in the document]
otherview: Introduced the EventWrapper which allows the user to check if it's data or there was an error with the event 
 or connection.

thorclient/common/common.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/vechain/thor/v2/thor"
+)
+
+const (
+	FinalizedRevision = "finalized"
+)
+
+var (
+	NotFoundErr      = fmt.Errorf("not found")
```
darrenvechain, line 19:
```suggestion
	ErrNotFound      = fmt.Errorf("not found")
```

thorclient/common/common.go
```diff
@@ -0,0 +1,32 @@
...
+	Not200StatusErr  = fmt.Errorf("not 200 status code")
```
darrenvechain, line 20:
```suggestion
	ErrStatus200  = fmt.Errorf("not 200 status code")
```

thorclient/common/common.go
```diff
@@ -0,0 +1,32 @@
...
+	UnexpectedMsgErr = fmt.Errorf("unexpected message format")
```
darrenvechain, line 21:
```suggestion
	ErrUnexpectedMsg = fmt.Errorf("unexpected message format")
```

thorclient/common/common.go
```diff
@@ -0,0 +1,32 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package common
```
darrenvechain, line 6: Does this need its own package? Seems tiny and could easily be placed inline with httpclient or thorclient

thorclient/httpclient/client.go
```diff
@@ -0,0 +1,221 @@
...
+	"github.com/vechain/thor/v2/thor"
+	"github.com/vechain/thor/v2/thorclient/common"
+)
+
+type Client struct {
+	url string
+	c   *http.Client
+}
+
+func NewClient(url string) *Client {
```
darrenvechain, line 29:
```suggestion
func New(url string) *Client {
```

thorclient/thorclient.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (c *Client) GetBlock(block string) (blocks *blocks.JSONBlockSummary, err error) {
+	return c.httpConn.GetBlock(block)
+}
+
+func (c *Client) GetTransaction(id *thor.Bytes32) (*transactions.Transaction, error) {
+	return c.httpConn.GetTransaction(id, false)
+}
+
+func (c *Client) GetTransactionPending(id thor.Bytes32) (*transactions.Transaction, error) {
```
darrenvechain, line 124: We have inconsistent parameters.

Pointer: 
```golang
GetTransaction(id *thor.Bytes32)
```


Non pointer:
```golang
GetTransactionPending(id thor.Bytes32)
```

`transaction.ID()` returns a non pointer, so I would be inclined to use that:

```golang
func (t *Transaction) ID() (id thor.Bytes32)
```

thorclient/httpclient/http_request.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/vechain/thor/v2/thorclient/common"
+)
+
+func (c *Client) httpRequest(method, url string, payload io.Reader) ([]byte, error) {
+	body, statusCode, err := c.rawHTTPRequest(method, url, payload)
+	if err != nil {
+		return nil, err
+	}
+	if statusCode != http.StatusOK {
```
darrenvechain, line 23: Should this be `2xx` instead of `StatusOK` ?

thorclient/common/common.go
otherview: Fixed 👍

thorclient/common/common.go
otherview: Fixed 👍

thorclient/common/common.go
otherview: Fixed 👍

thorclient/common/common.go
otherview: It's imported by multiple packages. Errors for error testing, `TxSendResult` for tx results and the EventWrapper for other uses.

But it's a good point, what if I leave this comment open and review it again once the PR is closer to done ?

thorclient/httpclient/client.go
otherview: Fixed 👍

thorclient/thorclient.go
otherview: Good point, fixed ! 

> transaction.ID() returns a non pointer, so I would be inclined to use that:
> func (t *Transaction) ID() (id thor.Bytes32)

I think the returning `transactions.Transaction` doesn't have that method, but it does have the ID field. Given that's it's lighter to pass around pointer, it's probably worth defaulting to `id *thor.Bytes32`? Wdyt ?

thorclient/httpclient/http_request.go
otherview: Does the api return other status codes for ok other than 200 ?

thorclient/httpclient/http_request.go
```diff
@@ -0,0 +1,77 @@
...
+		data = payload.([]byte)
+	} else {
+		data, err = json.Marshal(payload)
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to unmarshal payload - %w", err)
+	}
+
+	if string(data) == "[]" {
```
darrenvechain, line 72: `[]` is a valid response IMO, no need to error here.

If I query logs that don't have matching events an empty array is still a valid response

otherview: Yeah, but this is an http request, I think the reason this is here is that our api doesn't allow to receive empty lists.

api/blocks/blocks_test.go
```diff
@@ -87,7 +97,8 @@ func testGetBlockByHeight(t *testing.T) {
 }
 
 func testGetFinalizedBlock(t *testing.T) {
-	res, statusCode := httpGet(t, ts.URL+"/blocks/finalized")
+	res, statusCode, err := tclient.RawHTTPClient().RawHTTPGet("/blocks/finalized")
```
leszek-vechain, line 100: possibly we can add test for justified block

libotony:
For the import path:

1. `github.com/vechain/thor/thorclient`
2. `github.com/vechain/thor/client`

I would prefer option 2, which is simple, wdyt? @paologalligit @darrenvechain @otherview

otherview:
> For the import path:
> 
> 1. `github.com/vechain/thor/thorclient`
> 2. `github.com/vechain/thor/client`
> 
> I would prefer option 2, which is simple, wdyt? @paologalligit @darrenvechain @otherview

I'd be more inclined to `thorclient` as it's more distinguishable and similar to `ethclient`.

paologalligit:
> For the import path:
> 
> 1. `github.com/vechain/thor/thorclient`
> 2. `github.com/vechain/thor/client`
> 
> I would prefer option 2, which is simple, wdyt? @paologalligit @darrenvechain @otherview

If I have to choose between the two, I'd go with `thorclient`, but I'm open to a different not-too-generic naming

api/blocks/blocks_test.go
otherview: Yeah make sense to have that test, but perhaps it would be easier to introduce that test on the justified PR

thorclient/httpclient/client.go
```diff
@@ -0,0 +1,280 @@
...
+
+	var block blocks.JSONExpandedBlock
+	if err = json.Unmarshal(body, &block); err != nil {
+		return nil, fmt.Errorf("unable to unmarshall events - %w", err)
+	}
+
+	return &block, nil
+}
+
+func (c *Client) GetBlock(blockID string) (*blocks.JSONBlockSummary, error) {
```
darrenvechain, line 190: `JSONBlockSummary` is the wrong type

otherview: Good catch, it's the `JSONCollapsedBlock`. Fixed 👍

darrenvechain (approved):

thorclient/thorclient.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (c *Client) ChainTag() (byte, error) {
+	genesisBlock, err := c.Block("0")
+	if err != nil {
+		return 0, err
+	}
+	return genesisBlock.ID[31], nil
+}
+
+func (c *Client) SubscribeBlocks() (blocks <-chan common.EventWrapper[*blocks.JSONBlockSummary], err error) {
```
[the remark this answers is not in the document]
otherview: Fixing the tests and subs, seems to be used out everywhere 🦝

libotony:
> For the import path:
> 
> 1. `github.com/vechain/thor/thorclient`
> 2. `github.com/vechain/thor/client`
> 
> I would prefer option 2, which is simple, wdyt? @paologalligit @darrenvechain @otherview

Not a fan of `github.com/vechain/thor/thorclient` but since you guys are more into it then I am ok. And v2 kind of saved me, the import path is `github.com/vechain/thor/v2/thorclient`, a little bit not that anoy anymore.

api/transactions/types.go
```diff
@@ -220,3 +220,8 @@ func convertReceipt(txReceipt *tx.Receipt, header *block.Header, tx *tx.Transact
 	}
 	return receipt, nil
 }
+
+// TxSendResult is the response to the Send Tx method
+type TxSendResult struct {
```
libotony, line 225:
```suggestion
type SendTxResult struct {
```
Following the name convention of `GetCodeResult`.

thorclient/httpclient/client.go
```diff
@@ -0,0 +1,280 @@
...
+	return &Client{
+		url: url,
+		c:   &http.Client{},
+	}
+}
+
+func (c *Client) GetTransactionReceipt(txID *thor.Bytes32, revision string) (*transactions.Receipt, error) {
+	url := c.url + "/transactions/" + txID.String() + "/receipt"
+	if revision != "" {
+		url += "?revision=" + revision
```
libotony, line 39: Should be `head` other than `revision`.

thorclient/common/common.go
```diff
@@ -0,0 +1,27 @@
...
+	"fmt"
+)
+
+const (
+	BestRevision      = "best"
+	FinalizedRevision = "finalized"
+)
+
+var (
+	ErrNotFound      = fmt.Errorf("not found")
```
libotony, line 18:
```suggestion
	ErrNotFound      = errors.New("not found") 
```
Correct ?

thorclient/httpclient/client.go
otherview: Fixed 👍

api/transactions/types.go
otherview: Fixed !

thorclient/common/common.go
otherview: Yep!

thorclient/httpclient/client.go
```diff
@@ -0,0 +1,280 @@
...
+	return &block, nil
+}
+
+func (c *Client) GetBlock(blockID string) (*blocks.JSONCollapsedBlock, error) {
+	body, err := c.httpGET(c.url + "/blocks/" + blockID)
+	if err != nil {
+		return nil, fmt.Errorf("unable to retrieve block - %w", err)
+	}
+
+	var block blocks.JSONCollapsedBlock
```
libotony, line 196: Handle the scenario that returns `[]byte{"null"}`

thorclient/httpclient/client.go
```diff
@@ -0,0 +1,280 @@
...
+
+	var res accounts.GetStorageResult
+	if err = json.Unmarshal(body, &res); err != nil {
+		return nil, fmt.Errorf("unable to unmarshall code - %w", err)
+	}
+
+	return &res, nil
+}
+
+func (c *Client) GetBlockExpanded(revision string) (*blocks.JSONExpandedBlock, error) {
```
libotony, line 176:
```suggestion
func (c *Client) GetExpandedBlock(revision string) (*blocks.JSONExpandedBlock, error) {
```

thorclient/thorclient.go
```diff
@@ -0,0 +1,234 @@
...
+	options := applyOptions(opts)
+	return c.httpConn.GetTransactionReceipt(id, options.revision)
+}
+
+func (c *Client) InspectClauses(calldata *accounts.BatchCallData, opts ...Option) ([]*accounts.CallResult, error) {
+	options := applyOptions(opts)
+	return c.httpConn.InspectClauses(calldata, options.revision)
+}
+
+func (c *Client) InspectTxClauses(tx *tx.Transaction, senderAddr *thor.Address, opts ...Option) ([]*accounts.CallResult, error) {
```
libotony, line 100:
```suggestion
// accepts both signed Tx and unsigned Tx
func (c *Client) InspectTxClauses(tx *tx.Transaction, senderAddr *thor.Address, opts ...Option) ([]*accounts.CallResult, error) {
```

thorclient/httpclient/http_request.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (c *Client) rawHTTPRequest(method, url string, payload io.Reader) ([]byte, int, error) {
+	req, err := http.NewRequest(method, url, payload)
+	if err != nil {
+		return nil, 0, fmt.Errorf("error creating request: %w", err)
+	}
+
+	if method == "POST" {
+		req.Header.Set("Content-Type", "application/json")
+		req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
```
libotony, line 37: Do we need `application/x-www-form-urlencoded`?

thorclient/httpclient/http_request.go
libotony: What about `{}`?

thorclient/wsclient/client.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to connect - %w", err)
+	}
+
+	return subscribe[subscriptions.Beat2Message](conn)
+}
+
+// subscribe creates a channel to handle new subscriptions
+// It takes a websocket connection as an argument and returns a read-only channel for receiving messages of type T and an error if any occurs.
+func subscribe[T any](conn *websocket.Conn) (<-chan common.EventWrapper[*T], error) {
```
libotony, line 91: Looks like there is no close func for subscriptions, is it intended?

otherview: Added a `Subscriptions` that handles the closing of the websockets, lmk what you think!

thorclient/httpclient/http_request.go
otherview: Fixed 👍

thorclient/thorclient.go
otherview: Added !

thorclient/httpclient/client.go
otherview: Fixed !

thorclient/httpclient/client.go
otherview: Fixed 👍

libotony:
@otherview Appended a commit https://github.com/vechain/thor/pull/818/commits/0ec320d5242f9032e63ef114a19431112cca6734, adjusted the orders of the function to match the API doc.

thorclient/wsclient/client.go
libotony: I think it might be over engineering? `co.Goes` is for managing sync group, the close function is better managed by the caller.  I think the close function can be easily done in-place in `subscribe` which is clearer.

Another point is the producer won't need to consume all the remaining messages, it's consumer's decision to make. And it is a race condition for both consuming one channel. The producer should right away stop producing new events after receiving the stop signal.

otherview: I understand. I'm happy with going with another route, let's chat about it offline, but here are the reasoning behind those choices:

- `Subscriptions` as an object
Serves to encapsulate the `Unsubscribe` function (and potentially encapsulate an error channel if required)
I think it's neater to return 1 object and error :
`SubscribeBlocks() (*common.Subscription[*blocks.JSONCollapsedBlock], error)`
vs to return 1 object, one function that closes out the object and error:
`SubscribeBlocks() (*common.EventChan[*blocks.JSONCollapsedBlock], func() error, error)`

- `co.Goes` does not have a stop functionality. `co.Choes` adds a channel that allows to stop all goroutines.
When the `Unsubscribe` function is called, the `co.Choes.stopChan` is signalled to stop execution, making the subscription clean exit.
The caller still has to implement how it listens to the `co.Choes.stopChan` and exits its execution.

- I see what you mean. The comments on the drain messages are not on-point. It's actually only reading one (perhaps a few) event so that the `case <-stopChan:` is picked up. 
Once the `stopChan` is closed, the `eventChan` and the `conn` are closed, and no more messages are read.

What happened before "drainage" was that `err := conn.ReadJSON(&data)` will lock the routine, and it would never reach `case <-stopChan:`. 
I've seen this pattern happen quite a lot because `conn.ReadJSON` is locking.

libotony:
@otherview Appended a commit https://github.com/vechain/thor/pull/818/commits/a2569e0f0db4d8a6f8c96969f8f2c927edc2d2f6

thorclient/wsclient/client.go
```diff
@@ -0,0 +1,172 @@
...
+			eventChan <- common.EventWrapper[*T]{Data: &data}
+		}
+	}()
+
+	return &common.Subscription[*T]{
+		EventChan: eventChan,
+		Unsubscribe: func() {
+			closed = true
+			conn.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
+			conn.Close()
```
libotony, line 144: `// todo add metrics` was removed since this is a client code, mostly not running with in thor program.

thorclient/wsclient/client.go
```diff
@@ -0,0 +1,172 @@
...
+		return nil, err
+	}
+
+	conn.SetPingHandler(func(payload string) error {
+		// Make a best effort to send the pong message.
+		_ = conn.WriteControl(websocket.PongMessage, []byte(payload), time.Now().Add(time.Second))
+		conn.SetReadDeadline(time.Now().Add(readTimeout))
+		return nil
+	})
+	// TODO append to the connection pool
```
libotony, line 170: Not sure this comment is possible to achieve.

thorclient/thorclient.go
```diff
@@ -0,0 +1,274 @@
...
+// SubscribeTransfers subscribes to transfer updates over WebSocket.
+func (c *Client) SubscribeTransfers() (*common.Subscription[*subscriptions.TransferMessage], error) {
+	if c.wsConn == nil {
+		return nil, fmt.Errorf("not a websocket typed client")
+	}
+	return c.wsConn.SubscribeTransfers("")
+}
+
+// SubscribeBeats2 subscribes to Beat2 message updates over WebSocket.
+func (c *Client) SubscribeBeats2() (*common.Subscription[*subscriptions.Beat2Message], error) {
```
libotony, line 229: Query `pos` is required for all subscriptions.

thorclient/thorclient.go
```diff
@@ -0,0 +1,274 @@
...
+// SubscribeBlocks subscribes to block updates over WebSocket.
+func (c *Client) SubscribeBlocks() (*common.Subscription[*blocks.JSONCollapsedBlock], error) {
+	if c.wsConn == nil {
+		return nil, fmt.Errorf("not a websocket typed client")
+	}
+	return c.wsConn.SubscribeBlocks("")
+}
+
+// SubscribeEvents subscribes to event updates over WebSocket.
+func (c *Client) SubscribeEvents() (*common.Subscription[*subscriptions.EventMessage], error) {
```
libotony, line 213: Optional arguments https://mainnet.vechain.org/doc/stoplight-ui/#/paths/subscriptions-transfer/get, same for transfers.

thorclient/wsclient/client.go
otherview: We should have a limited connection pool in the future. 
Also, we should re-use the websocket connection instead creating one per request, although that might be tricky, impossible given we're connecting sockets to specific endpoints.

thorclient/thorclient.go
otherview: Good spot. Fixed 👍

thorclient/thorclient.go
otherview: Good spot. Fixed 👍

thorclient/thorclient.go
```diff
@@ -0,0 +1,274 @@
...
+	return c.httpConn.GetExpandedBlock(revision)
+}
+
+// FilterEvents filters events based on the provided filter request.
+func (c *Client) FilterEvents(req *events.EventFilter) ([]events.FilteredEvent, error) {
+	return c.httpConn.FilterEvents(req)
+}
+
+// FilterTransfers filters transfers based on the provided filter request.
+func (c *Client) FilterTransfers(req *events.EventFilter) ([]*transfers.FilteredTransfer, error) {
```
libotony, line 186: Should be `TransferFilter`

paologalligit: We already have `filter - transfer` style in the code, like [here](https://github.com/vechain/thor/blob/master/api/transfers/transfers.go#L58), also the returned data structure is called `FilteredTransfer`. I would keep the naming consistent, wdyt?

libotony: Sorry for the confusion, the parameter in the function. Should be `github.com/vechain/thor/api/transfers.TransferFilter`

paologalligit: Ohh, you're right, good catch!

libotony (approved):

PR #821, opened 2024-08-16
chore: add a histogram for tx coefficients

darrenvechain:
# Description

Add a histogram metric for tx coefficients

# Checklist:


consensus/validator.go
```diff
@@ -328,6 +328,7 @@ func (c *Consensus) verifyBlock(blk *block.Block, state *state.State, blockConfl
 		totalGasUsed += receipt.GasUsed
 		receipts = append(receipts, receipt)
 		processedTxs[tx.ID()] = receipt.Reverted
+		metricCoefBucket().Observe(int64(tx.GasPriceCoef()))
```
otherview, line 331: What happens if this block is invalid ?
What if this metric was set somewhere near the `metricBlockProcessedxxx` metrics ?

darrenvechain: Fixed

libotony:
IMO, this one is better collected off chain, like https://github.com/vechain/vestatus

libotony:
> > IMO, this one is better collected off chain, like https://github.com/vechain/vestatus
> 
> Shouldn't thor allow all data that can be collected from the node to be available ? I think this falls similar to the block gas and tx gas metrics.

I think the scope of thor metrics should be limited to the runtime(real time) info. Block/Tx related is better collect off the client, the `time` of the 'facts' related metrics was born with it(block time) rather than the time it was collected.

darrenvechain:
Closing as we can do this with historical data

PR #823, opened 2024-08-19
Add api-allowed-tracers flag

otherview:
# Description

This is a tentative PR with regard to https://github.com/vechain/protocol-board-repo/issues/292

- Adds a new `allowed-tracers` flag. Ex: `--allowed-tracers="all"`, `--allowed-tracers="call,4byte"
- The default available tracers are changed - `--allowed-tracers="none"` is the default
- Node operator can now pick-and-choose which tracers are available
- "" is no longer accepted as a valid (defaulted to log) tracer name

Fixes # https://github.com/vechain/protocol-board-repo/issues/292

## Type of change

# How Has This Been Tested?

- [x] Tests

# Checklist:

### DIFF
--- a/api/debug/debug_test.go
+++ b/api/debug/debug_test.go
@@ -53,6 +53,7 @@ func TestDebug(t *testing.T) {
 	// /tracers endpoint
 	for name, tt := range map[string]func(*testing.T){
 		"testTraceClauseWithEmptyTracerName":       testTraceClauseWithEmptyTracerName,
+		"testTraceClauseWithInvalidTracerName":     testTraceClauseWithInvalidTracerName,
 		"testTraceClauseWithEmptyTracerTarget":     testTraceClauseWithEmptyTracerTarget,
 		"testTraceClauseWithBadBlockId":            testTraceClauseWithBadBlockId,
 		"testTraceClauseWithNonExistingBlockId":    testTraceClauseWithNonExistingBlockId,
@@ -163,6 +164,11 @@ func testTraceClauseWithEmptyTracerName(t *testing.T) {
 
 	res = httpPostAndCheckResponseStatus(t, ts.URL+"/debug/tracers", &TraceClauseOption{Name: " "}, 403)
 	assert.Equal(t, "tracer name must be defined", strings.TrimSpace(res))
+}
+
+func testTraceClauseWithInvalidTracerName(t *testing.T) {
+	res := httpPostAndCheckResponseStatus(t, ts.URL+"/debug/tracers", &TraceClauseOption{Name: "non-existent"}, 403)
+	assert.Contains(t, res, "unable to create custom tracer")
 }
 
 func testTraceClauseWithEmptyTracerTarget(t *testing.T) {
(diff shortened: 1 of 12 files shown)


cmd/thor/flags.go
```diff
@@ -151,6 +151,12 @@ var (
 		Usage: "metrics service listening address",
 	}
 
+	allowedTracersFlag = cli.StringFlag{
+		Name:  "allowed-tracers",
```
darrenvechain, line 155:
```suggestion
		Name:  "api-allowed-tracers",
```

@otherview to keep consistency with `api-allow-custom-tracers`

darrenvechain: Also, update the usage.md if you can

cmd/thor/utils.go
```diff
@@ -754,3 +754,16 @@ func readIntFromUInt64Flag(val uint64) (int, error) {
 
 	return i, nil
 }
+
+func parseTracerList(list string) map[string]interface{} {
```
darrenvechain, line 758: Should we validate inputs? As in if I start thor with `thor --allowed-tracers="darren,123"`

cmd/thor/flags.go
```diff
@@ -151,6 +151,12 @@ var (
...
+		Value: "all",
```
darrenvechain, line 156: Should this be `"*"` to be consistent with cors?

cmd/thor/flags.go
otherview: Good point - fixed !

cmd/thor/flags.go
otherview: Doesn't matter I think.. I believe that in CORS the `*` is a regex parser.
In the `api-allowed-tracers` it's just a keyword.
Wdyt ?

cmd/thor/utils.go
otherview: If someone tries to use the `123` tracer it will fail because it's not loaded. I think it should be ok ?
We could have a list of available tracers, but I didn't want to change too much the existing debug package.

api/debug/debug_test.go
```diff
@@ -156,13 +157,22 @@ func TestStorageRangeMaxResult(t *testing.T) {
 	assert.Equal(t, 10, len(storageRangeRes.Storage))
 }
 
+func testTraceClauseWithEmptyTracerName(t *testing.T) {
```
leszek-vechain, line 160: do we have any test for invalid tracer name ?

cmd/thor/utils.go
leszek-vechain: but what about darren tracer ? I think maybe it is worth validating it

api/debug/debug_test.go
otherview: Added !

cmd/thor/utils.go
otherview: I believe that if the node operator is specifying a tracer by name, he/she should double-check that the name is correct. Happy to iterate on this feature in the future if there's feedback in that direction.

darrenvechain (dismissed):

darrenvechain (approved):

sara554-beep (changes requested):
@ver

PR #825, opened 2024-08-21
Refactor metrics and admin servers

MakisChristou:
This is a draft PR regarding the refactoring of Metrics and Admin servers as well as adding tests to Admin.


admin/admin_test.go
```diff
@@ -0,0 +1,101 @@
...
+	"bytes"
+	"encoding/json"
+	"log/slog"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+func TestPostLogLevelHandler_ValidInput(t *testing.T) {
+	var logLevel slog.LevelVar
```
otherview, line 18: Good tests! I wonder if they can be packed into a table test ?

admin/admin.go
```diff
@@ -6,83 +6,17 @@
 package admin
```
otherview, line 6: Nit: Perhaps `admin_server.go` `admin.go` ? To follow the `metrics_server.go` convention ?

otherview (dismissed):
lgtm, minor bits worth chatting and perhaps changing but nothing blocking the merge imo.

libotony:
Considering the code structure, I would consider moving admin folder into api folder and a `admin_server.go` in the API root folder. @MakisChristou wdyt?

otherview (approved):
lgtm - Would perhaps put the `api/admin/admin_server.go` into `api/admin_server.go` so that's at the same level as the `metrics_server.go` other than that lgtm.

Wdyt @libotony ?

otherview (dismissed):

libotony:
Not part of this PR but related to the admin component @MakisChristou 

``` go
func writeError(w http.ResponseWriter, errCode int, errMsg string) {
	w.Header().Set("Content-Type", "application/json")
	w.WriteHeader(errCode)
	json.NewEncoder(w).Encode(errorResponse{
		ErrorCode:    errCode,
		ErrorMessage: errMsg,
	})
}
```
Setting an HTTP status code and also including the code in the JSON response appears to be redundant.

libotony:
> lgtm - Would perhaps put the `api/admin/admin_server.go` into `api/admin_server.go` so that's at the same level as the `metrics_server.go` other than that lgtm.
> 
> Wdyt @libotony ?

Sounds good to me 👍 .

libotony (approved):

PR #832, opened 2024-08-28
Adding transaction_call endpoint

otherview:
# Description

This PR adds a new endpoint that allows to simulate and estimate based on transactions instead of clauses.
The added value is to bring api communication to the transaction level. This allows for gotchas like, chaintag, expiration, nonce, etc to be more visible and allows executing transactions in a given block revision

Summary of the endpoint:
```
URL :         POST /transactions/call?revision=best
Payload :   transaction.Transaction (existing type)
Response:   transaction.CallReceipt (transaction.Receipt minus Block info)
type CallReceipt struct {
	GasUsed  uint64                `json:"gasUsed"`
	GasPayer thor.Address          `json:"gasPayer"`
	Paid     *math.HexOrDecimal256 `json:"paid"`
	Reward   *math.HexOrDecimal256 `json:"reward"`
	Reverted bool                  `json:"reverted"`
	TxID     thor.Bytes32          `json:"txID"`
	TxOrigin thor.Address          `json:"txOrigin"`
	Outputs  []*Output             `json:"outputs"`
	VmError  string                `json:"vmError"`
}
```
Goal: Execute a transaction at the head of a specified block. Retrieve the total gas cost at the transaction level.
(Note: Can be expanded to execute in a block at a given tx position.)

Fixes # [(issue)](https://github.com/vechain/protocol-board-repo/issues/157)

## Type of change

# How Has This Been Tested?
- [x] Tests

# Checklist:


thorclient/httpclient/client_test.go
```diff
@@ -300,6 +301,25 @@ func TestClient_GetTransaction(t *testing.T) {
 	assert.Equal(t, expectedTx, tx)
 }
 
+func TestClient_CallTransaction(t *testing.T) {
```
leszek-vechain, line 304: can we also add negative tests ?

api/transactions/transactions.go
```diff
@@ -213,6 +228,85 @@ func (t *Transactions) parseHead(head string) (thor.Bytes32, error) {
 	return h, nil
 }
 
+func (t *Transactions) txCall(
+	txCallMsg *Transaction,
+	header *block.Header,
+	st *state.State,
+) (*CallReceipt, error) {
+	callAddr := txCallMsg.Origin
+	if callAddr.String() == (thor.Address{}).String() {
```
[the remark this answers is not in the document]
MakisChristou: We do don't we?

```golang
{
	testTx: transactions.ConvertCoreTransaction(new(tx.Builder).
		ChainTag(chainTag).
		Gas(gas).
		Build(),
		nil, &thor.Address{}, nil),
	errMsg: "no origin address specified",
},
```

thorclient/httpclient/client_test.go
MakisChristou: I guess we can but those are mock tests afaik. So not sure what the benefit of returning a hardcoded "error" response will be and checking if its what we expect.

api/doc/thor.yaml
```diff
@@ -191,6 +191,177 @@ paths:
                 type: string
                 example: 'Invalid transaction ID'
 
+  /transactions/call:
```
otherview, line 194: I haven't tried but when this yml gets generated, does it contain a working example ? (like the other endpoints)

api/metrics.go
```diff
@@ -71,10 +97,23 @@ func metricsMiddleware(next http.Handler) http.Handler {
...
 		if rt != nil && rt.GetName() != "" {
 			enabled = true
 			name = rt.GetName()
+			if name == "transactions_call_tx" {
+				ctxWriter := newCallTxResponseWriter(w)
+				next.ServeHTTP(ctxWriter, r)
+
+				// Record VM error if present
+				if ctxWriter.VMError != "" {
+					metricTxCallVMErrors().AddWithLabel(1, map[string]string{
```
otherview, line 109: What happens if there are a high number of different errors here ?

api/metrics.go
```diff
@@ -71,10 +97,23 @@ func metricsMiddleware(next http.Handler) http.Handler {
 			subscription = false
 		)
 
-		// all named route will be recorded
 		if rt != nil && rt.GetName() != "" {
 			enabled = true
 			name = rt.GetName()
+			if name == "transactions_call_tx" {
```
otherview, line 103: What is the goal of this particular case ?

api/transactions/transactions_test.go
```diff
@@ -74,6 +79,16 @@ func TestTransaction(t *testing.T) {
 	} {
 		t.Run(name, tt)
 	}
+
+	// Call transaction
+	for name, tt := range map[string]func(*testing.T){
```
otherview, line 84: Can another test be added here that does the same checks, or a summarized version, of the checks that the mempool does to verify tx integrity ?

otherview: Ah! These seem to be covered already yeah ?

otherview (commented):
Very minor bits, but lgtm - should there be a couple of e2e to go with this as well ?
I can't approve, but perhaps @libotony can have a final review and approve ?

api/doc/thor.yaml
YeahNotSewerSide: I think so, but do you mean examples for values? like [here](https://github.com/vechain/thor/blob/cb0e53dbd954f89d9e1b1744d10d538e75394939/api/doc/thor.yaml#L214)
or code/usage example?

api/metrics.go
YeahNotSewerSide: I guess to separate errors that happened specifically on that endpoint, the name for it is set [here](https://github.com/vechain/thor/blob/cb0e53dbd954f89d9e1b1744d10d538e75394939/api/transactions/transactions.go#L327). Not sure if that metric is needed, since it's the errors that happened on transactions, that are not mined

api/metrics.go
YeahNotSewerSide: From the code I read, there is always at most only 1 possible vmerror per transaction, that doesn't depend on the amount of clauses in a transaction. Was that the question?

libotony:
Discussed offline, @otherview pointed that including runtime features is potentially providing values for devs, I would like to have a further investigation into the purpose of this PR.

api/transactions/transactions_test.go
```diff
@@ -257,6 +270,163 @@ func handleGetTransactionReceiptByIDWithNonExistingHead(t *testing.T) {
...
+			errMsg: "no origin address specified",
+		},
+		{
+			testTx: transactions.ConvertCoreTransaction(new(tx.Builder).
+				ChainTag(chainTag).
+				Gas(gas).
+				Clause(tx.NewClause(nil).WithData(make([]byte, 64*1024+1))).
+				Build(),
+				nil, sendAddr, nil),
+			errMsg: "size too large",
```
leszek-vechain, line 394: wonder if we can add test for features being invalid (as we have error check in transactions.go)

YeahNotSewerSide: @leszek-vechain I reviewed the code again, and this [check](https://github.com/vechain/thor/blob/65005a263919a77b680adf5c77a6a02d9b23d493/api/transactions/transactions.go#L253) is unreachable, because of [this](https://github.com/vechain/thor/blob/65005a263919a77b680adf5c77a6a02d9b23d493/api/transactions/types.go#L336) which will be always setting features to the correct features for the block, not sure if that's an intended behavior or not

YeahNotSewerSide: From what I understand this check is entirely redundant

YeahNotSewerSide: removed it

PR #836, opened 2024-09-04
Adding Health endpoint

otherview:
# Description

A health check service that lives inside the node as an API endpoint.

I think it's a good illustrative start for a PR, there are a few tech choices made that I consider would be worth discussing, such as:
- Should this endpoint live in the Admin API ?
- Should this be a singleton service ?
- Is the naming on point ?

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/cmd/thor/main.go
+++ b/cmd/thor/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/vechain/thor/v2/cmd/thor/optimizer"
 	"github.com/vechain/thor/v2/cmd/thor/solo"
 	"github.com/vechain/thor/v2/genesis"
+	"github.com/vechain/thor/v2/health"
 	"github.com/vechain/thor/v2/log"
 	"github.com/vechain/thor/v2/logdb"
 	"github.com/vechain/thor/v2/metrics"
@@ -222,6 +223,7 @@ func defaultAction(ctx *cli.Context) error {
 		return err
 	}
 
+	healthStatus := &health.Health{}
 	printStartupMessage1(gene, repo, master, instanceDir, forkConfig)
 
 	if !skipLogs {
@@ -238,7 +240,7 @@ func defaultAction(ctx *cli.Context) error {
 	txPool := txpool.New(repo, state.NewStater(mainDB), txpoolOpt)
 	defer func() { log.Info("closing tx pool..."); txPool.Close() }()
 
-	p2pCommunicator, err := newP2PCommunicator(ctx, repo, txPool, instanceDir)
+	p2pCommunicator, err := newP2PCommunicator(ctx, repo, txPool, instanceDir, healthStatus)
 	if err != nil {
 		return err
 	}
@@ -255,6 +257,7 @@ func defaultAction(ctx *cli.Context) error {
 		logDB,
 		bftEngine,
 		p2pCommunicator.Communicator(),
+		healthStatus,
 		forkConfig,
 		ctx.String(apiCorsFlag.Name),
 		uint32(ctx.Uint64(apiBacktraceLimitFlag.Name)),
@@ -297,7 +300,9 @@ func defaultAction(ctx *cli.Context) error {
 		p2pCommunicator.Communicator(),
 		ctx.Uint64(targetGasLimitFlag.Name),
 		skipLogs,
-		forkConfig).Run(exitSignal)
+		forkConfig,
+		healthStatus,
+	).Run(exitSignal)
 }
 
 func soloAction(ctx *cli.Context) error {
@@ -399,6 +404,8 @@ func soloAction(ctx *cli.Context) error {
 	defer func() { log.Info("closing tx pool..."); txPool.Close() }()
 
 	bftEngine := solo.NewBFTEngine(repo)
+	healthStatus := &health.Health{}
+
 	apiHandler, apiCloser := api.New(
 		repo,
 		state.NewStater(mainDB),
@@ -406,6 +413,7 @@ func soloAction(ctx *cli.Context) error {
 		logDB,
 		bftEngine,
 		&solo.Communicator{},
+		healthStatus,
 		forkConfig,
 		ctx.String(apiCorsFlag.Name),
 		uint32(ctx.Uint64(apiBacktraceLimitFlag.Name)),
@@ -443,6 +451,7 @@ func soloAction(ctx *cli.Context) error {
 	return solo.New(repo,
 		state.NewStater(mainDB),
 		logDB,
+		healthStatus,
 		txPool,
 		ctx.Uint64(gasLimitFlag.Name),
 		ctx.Bool(onDemandFlag.Name),
--- a/health/health.go
+++ b/health/health.go
@@ -1,0 +1,66 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package health
+
+import (
+	"sync"
+	"time"
+
+	"github.com/vechain/thor/v2/thor"
+)
+
+type BlockIngestion struct {
+	BestBlock          *thor.Bytes32 `json:"bestBlock"`
+	BestBlockTimestamp *time.Time    `json:"bestBlockTimestamp"`
+}
+
+type Status struct {
+	Healthy        bool            `json:"healthy"`
+	BlockIngestion *BlockIngestion `json:"blockIngestion"`
+	ChainSync      bool            `json:"chainSync"`
+}
+
+type Health struct {
+	lock         sync.RWMutex
+	newBestBlock time.Time
+	bestBlockID  *thor.Bytes32
+	chainSynced  bool
+}
+
+func (h *Health) NewBestBlock(ID thor.Bytes32) {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	h.newBestBlock = time.Now()
+	h.bestBlockID = &ID
+}
+
+func (h *Health) Status() (*Status, error) {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+
+	blockIngest := &BlockIngestion{
+		BestBlock:          h.bestBlockID,
+		BestBlockTimestamp: &h.newBestBlock,
+	}
+
+	// todo review time slots
+	healthy := time.Since(h.newBestBlock) >= 10*time.Second &&
+		h.chainSynced
+
+	return &Status{
+		Healthy:        healthy,
+		BlockIngestion: blockIngest,
+		ChainSync:      h.chainSynced,
+	}, nil
+}
+
+func (h *Health) ChainSynced() {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	h.chainSynced = true
+}
--- a/api/api.go
+++ b/api/api.go
@@ -17,6 +17,7 @@ import (
 	"github.com/vechain/thor/v2/api/debug"
 	"github.com/vechain/thor/v2/api/doc"
 	"github.com/vechain/thor/v2/api/events"
+	"github.com/vechain/thor/v2/api/health"
 	"github.com/vechain/thor/v2/api/node"
 	"github.com/vechain/thor/v2/api/subscriptions"
 	"github.com/vechain/thor/v2/api/transactions"
@@ -28,6 +29,8 @@ import (
 	"github.com/vechain/thor/v2/state"
 	"github.com/vechain/thor/v2/thor"
 	"github.com/vechain/thor/v2/txpool"
+
+	healthstatus "github.com/vechain/thor/v2/health"
 )
 
 var logger = log.WithContext("pkg", "api")
@@ -40,6 +43,7 @@ func New(
 	logDB *logdb.LogDB,
 	bft bft.Committer,
 	nw node.Network,
+	healthStatus *healthstatus.Health,
 	forkConfig thor.ForkConfig,
 	allowedOrigins string,
 	backtraceLimit uint32,
@@ -73,6 +77,8 @@ func New(
 
 	accounts.New(repo, stater, callGasLimit, forkConfig, bft).
 		Mount(router, "/accounts")
+
+	health.New(healthStatus).Mount(router, "/health")
 
 	if !skipLogs {
 		events.New(repo, logDB, logsLimit).
--- a/api/health/health.go
+++ b/api/health/health.go
@@ -1,0 +1,41 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package health
+
+import (
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/vechain/thor/v2/api/utils"
+	"github.com/vechain/thor/v2/health"
+)
+
+type Health struct {
+	healthStatus *health.Health
+}
+
+func New(healthStatus *health.Health) *Health {
+	return &Health{
+		healthStatus: healthStatus,
+	}
+}
+
+func (h *Health) handleGetHealth(w http.ResponseWriter, req *http.Request) error {
+	acc, err := h.healthStatus.Status()
+	if err != nil {
+		return err
+	}
+	return utils.WriteJSON(w, acc)
+}
+
+func (h *Health) Mount(root *mux.Router, pathPrefix string) {
+	sub := root.PathPrefix(pathPrefix).Subrouter()
+
+	sub.Path("/").
+		Methods(http.MethodGet).
+		Name("health").
+		HandlerFunc(utils.WrapHandlerFunc(h.handleGetHealth))
+}
--- a/api/health/types.go
+++ b/api/health/types.go
@@ -1,0 +1,10 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package health
+
+type Response struct {
+	Healthy bool `json:"healthy"`
+}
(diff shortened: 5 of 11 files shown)


health/health.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"sync"
+	"time"
+
+	"github.com/vechain/thor/v2/thor"
+)
+
+type BlockIngestion struct {
+	BestBlock          *thor.Bytes32 `json:"bestBlock"`
+	BestBlockTimestamp *time.Time    `json:"bestBlockTimestamp"`
```
paologalligit, line 17: This could be misunderstood by the best block's timestamps itself, while this timestamp is from when it gets processed. Maybe `BestBlockTimestamp` can be renamed to `BestBlockIngestionTime` or `...ProcessedTime`

health/health.go
```diff
@@ -0,0 +1,66 @@
...
+		BlockIngestion: blockIngest,
+		ChainSync:      h.chainSynced,
+	}, nil
+}
+
+func (h *Health) ChainSynced() {
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	h.chainSynced = true
```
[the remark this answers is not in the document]
otherview: Good point, I've modified the ChainSync to ChainSyncStatus and marked as unhealthy if it's syncing.

There is a concern that this may make the ChainSync Status to flicker from Sync to UnSync a bit too often.

health/health.go
```diff
@@ -0,0 +1,66 @@
...
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+
+	blockIngest := &BlockIngestion{
+		BestBlock:          h.bestBlockID,
+		BestBlockTimestamp: &h.newBestBlock,
+	}
+
+	// todo review time slots
+	healthy := time.Since(h.newBestBlock) >= 10*time.Second &&
```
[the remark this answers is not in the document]
otherview: Yeah you're right, that's why the 10sec is marked as a todo, what would make sense for a this time gap you think ?

health/health.go
otherview: Good point - fixed to `BestBlockIngestionTimestamp`

leszek-vechain:
do we have any tests for this new endpoint ?

health/health.go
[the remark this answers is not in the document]
paologalligit: Maybe we could introduce some kind of time delay before switching state; I wouldn't know exactly how much, but that would make the switch less sudden. Would that be feasible?

health/health.go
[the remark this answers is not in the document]
paologalligit: What I don't get is, if we just received a block and check for the healthy condition, the time will be less then 10 seconds ago, so the node will be considered as not healthy, but why?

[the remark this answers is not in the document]
otherview: Gotcha! Yes, the condition was in reverse. I've added a unit test to make sure the logic stays correct now :)

health/health.go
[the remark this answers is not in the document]
otherview: That's a good idea, will implement, want to try this out with testnet/mainnet first to see if there is a flickering.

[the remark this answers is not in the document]
otherview: It doesn't seem to flicker but it's not working as intended. I have a few ideas that I would like to run by you, will ping you tomorrow 👍

api/admin/admin.go
otherview, line 1: The admin endpoints were buckled in this package and reformatted to follow the same pattern as other endpoints

api/admin_server.go
otherview, line 1: Potentially, this can live in the init functions of the thor and thor-solo ?

darrenvechain:
I got this when I sync a new node, is it expected?

```bash
curl http://localhost:2113/admin/health
```

```json
{"healthy":false,"blockIngestion":{"bestBlock":"0x00014fe5a9e48a02268655e8206bb4cbcc95a1ae0d1f5f4360f697f02077c727","bestBlockIngestionTimestamp":"2024-11-04T10:49:16.06031Z"},"chainSync":false}
```

image

darrenvechain:
Is the response always 200? From my biased experienced with spring actuator, it returns 503 if some aspect of the application is unhealthy. This makes it easier for @kgapos health checks as he doesn't have to get the response body and parse and check etc. He can just check the response code

kgapos:
Yeah, @darrenvechain is right, it's going to be useless as far as the AWS ALB is concerned if it doesn't return a 5XX status code when the node is unhealthy. I understand this is unfortunate and counter intuitive, but unless you build it like that I have to implement some sort of wrapper, which is the exact thing we're trying to replace. Tagging @otherview for visibility.

otherview:
> I got this when I sync a new node, is it expected?
> Edit: I just saw previous comment. IMO it is healthy if its syncing like this

I understand what you mean, it's the regular and expected behaviour to sync to the lastest block. 
Perhaps it's possible to find another naming system, (ready comes to mind,) but the idea is that a node is not healthy to provide node operations if it's in the sync period.

This is the standard I've seen in other nodes, as it helps node operators to know when the node is 100% ready to process blockchain operations.

otherview:
> Yeah, @darrenvechain is right, it's going to be useless as far as the AWS ALB is concerned if it doesn't return a 5XX status code when the node is unhealthy. I understand this is unfortunate and counter intuitive, but unless you build it like that I have to implement some sort of wrapper, which is the exact thing we're trying to replace. Tagging @otherview for visibility.


Yeah I got you covered :) It's returning a 503 when `"healthy":false`

health/health.go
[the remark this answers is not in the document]
otherview: Modified the healthcheck to bootstrapStatus from chainSync .

health/health.go
```diff
@@ -0,0 +1,65 @@
...
+func (h *Health) Status() (*Status, error) {
+	h.lock.RLock()
+	defer h.lock.RUnlock()
+
+	blockIngest := &BlockIngestion{
+		BestBlock:                   h.bestBlockID,
+		BestBlockIngestionTimestamp: &h.newBestBlock,
+	}
+
+	healthy := time.Since(h.newBestBlock) <= 10*time.Second && // less than 10 secs have passed since a new block was received
```
darrenvechain, line 50: Will this fail if the node is not in a UTC timezone?

Also, should we give some buffer in case we miss slots?

Another possible solution could be to get all the Block IDs from the peers, and find something like:
- Highest AND
- most common

Also, not sure if this is a concern, but solo can have an interval not equal to 10 seconds. I think the `blockInterval` is passed into some structs

otherview: Added the healthcheck for solo mode !

Checking the blocks from peers is a more elegant solution, (it's actually how the bootstrap is working) but given that it would require a lot more changes I think it's possible to enhance this feature in the next iteration ?

health/health.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/vechain/thor/v2/thor"
+)
+
+type BlockIngestion struct {
+	BestBlock                   *thor.Bytes32 `json:"bestBlock"`
+	BestBlockIngestionTimestamp *time.Time    `json:"bestBlockIngestionTimestamp"`
```
darrenvechain, line 18: Maybe just `timestamp` ? bestBlockIngestionTimestamp is already wrapped inside blockingestion.

BestBlock could also just be `ID` single the outer struct name already has `Block`

health/health.go
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package health
```
darrenvechain, line 6: This this package need to be at the root level? Seems very small

libotony:
I would suggest to only leverage best block's timestamp in this function.

health/health.go
otherview: Sure, added a test for the solo 👍

health/health.go
otherview: Good point ! Making those changes.

health/health.go
```diff
@@ -0,0 +1,81 @@
...
+		h.bootstrapStatus
+
+	return &Status{
+		Healthy:           healthy,
+		BlockIngestion:    blockIngest,
+		ChainBootstrapped: h.bootstrapStatus,
+	}, nil
+}
+
+func (h *Health) NewBestBlock(ID thor.Bytes32) {
```
libotony, line 68: This can be replaced by `*repository.NewTicker()`.

health/health.go
```diff
@@ -0,0 +1,81 @@
...
+	h.lock.Lock()
+	defer h.lock.Unlock()
+
+	h.newBestBlock = time.Now()
+	h.bestBlockID = &ID
+}
+
+func (h *Health) BootstrapStatus(bootstrapStatus bool) {
```
libotony, line 76: This can be ignored by a check to `repo.BestBlock().Timestamp()`?

libotony: And the config of solo can also be ignored.

health/health.go
libotony: Had two comments that has the potential to put this package into the `api/admin` folder.

health/health.go
otherview: Not sure if I'm following, do you mean to replace the `health. NewBestBlock` by a ticker that checks if new blocks have arrived ?
Changing from `node.health.NewBestBlock ` to `health.node.CheckBlockTimestamp` every x seconds ?

health/health.go
otherview: Yeah, we can infer that if the bestblock timestamp is near the current timestamp the bootstrapping process has finished. However, I believe that kinda changes the health package from a push to a pull ?

It's a pull as it's now the health responsibility to trigger certain methods in components to identify if they are healthy. Versus the push where components tell health if they are healthy or not.

api/admin/health/health.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func New(healthStatus *health.Health) *Health {
+	return &Health{
+		healthStatus: healthStatus,
+	}
+}
+
+func (h *Health) handleGetHealth(w http.ResponseWriter, _ *http.Request) error {
+	acc, err := h.healthStatus.Status()
```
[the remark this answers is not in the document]
otherview: Yeah, interesting.. We can certainly make it configurable!
What case are you thinking of where this would be handy ?

health/health.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func New(repo *chain.Repository, p2p *comm.Communicator, timeBetweenBlocks time.Duration) *Health {
+	return &Health{
+		repo:              repo,
+		timeBetweenBlocks: timeBetweenBlocks + delayBuffer,
+		p2p:               p2p,
+	}
+}
```
otherview, line 51: I refactored the health service to accept components and work in a pull fashion.
I think it looks better, thanks guys 🙏

health/health.go
```diff
@@ -0,0 +1,99 @@
+// Copyright (c) 2024 The VeChainThor developers
+
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
+package health
```
darrenvechain, line 6: I think this package could move under `admin/health` since it is self contained now

otherview: Fixed!  I typically decouple service from api, but I agree that there's not a lot of value in this case.

api/admin/health/health.go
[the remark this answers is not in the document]
otherview: Added !

api/admin/health/health_api.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+
+	if queryMinPeerCount := query.Get("minPeerCount"); queryMinPeerCount != "" {
+		if parsed, err := strconv.Atoi(queryMinPeerCount); err == nil {
+			minPeerCount = parsed
+		}
+	}
+
+	acc, err := h.healthStatus.Status(maxTimeBetweenSlots, minPeerCount)
```
otherview, line 48: Added http query params so it's possible to access different tolerance in health constraints.

api/admin/health/health.go
```diff
@@ -0,0 +1,110 @@
...
+func New(repo *chain.Repository, p2p *comm.Communicator) *Health {
+	return &Health{
+		repo: repo,
+		p2p:  p2p,
+	}
+}
+
+// isNetworkProgressing checks if the network is producing new blocks within the allowed interval.
+func (h *Health) isNetworkProgressing(now time.Time, bestBlockTimestamp time.Time, maxTimeBetweenSlots time.Duration) bool {
+	return now.Sub(bestBlockTimestamp) <= maxTimeBetweenSlots
```
darrenvechain, line 50: Does this cater for non-UTC timezones?

api/admin/health/health_api.go
```diff
@@ -0,0 +1,68 @@
...
+func (h *API) handleGetHealth(w http.ResponseWriter, r *http.Request) error {
+	// Parse query parameters
+	query := r.URL.Query()
+
+	// Default to constants if query parameters are not provided
+	maxTimeBetweenSlots := defaultMaxTimeBetweenSlots
+	minPeerCount := defaultMinPeerCount
+
+	// Override with query parameters if they exist
+	if queryMaxTimeBetweenSlots := query.Get("maxTimeBetweenSlots"); queryMaxTimeBetweenSlots != "" {
```
darrenvechain, line 36: How about `blockTolerance` or something else. Most people don't get the concept of slots straight away

api/admin/loglevel/log_level_test.go
```diff
@@ -14,6 +14,8 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/gorilla/mux"
+
```
darrenvechain, line 18: nit: whitespace

darrenvechain (approved):

api/admin/loglevel/log_level_test.go
otherview: Fixed !

api/admin/health/health_api.go
otherview: Yeah, `blockTolerance` is a much better name. Fixed!

api/admin/health/health.go
otherview: What leads you to say that? It should consider, as it's a `time.Time` instance right ?

darrenvechain:
```
bestBlockTimestamp := time.Unix(int64(bestBlock.Header.Timestamp()), 0)
```

`bestBlock.Header.Timestamp()` will be the same value, where the node is in America or Asia, because the block timestamp is always UTC. `time.Now()` will differ, depending on whether the node is in America or Asia

otherview: Ah, I see what you mean.
The `time.Time` holds the timestamp and a location (which I believe it defaults to UTC), so when you're comparing time objects it take into account the location so to avoid timezone issues.

From the https://go.dev/src/time/time.go
```
// Each time has an associated [Location]. The methods [Time.Local], [Time.UTC], and Time.In return a
// Time with a specific Location. Changing the Location of a Time value with
// these methods does not change the actual instant it represents, only the time
// zone in which to interpret it.
```

In this case I think I see an issue because `bestBlockTimestamp` should have a default location of UTC which is correct. So when comparing `timeNowInAsia` with `bestBlockTimestamp` it's going to correctly compare because of it's underlying storage structure.

darrenvechain: But the block does not store a time.Time, it stores seconds. 

This does not specify the timezone, so thats why I'm confused:

```
time.Unix(int64(bestBlock.Header.Timestamp()), 0)
```

darrenvechain: Ahh, I got it, `Unix` is always UTC

darrenvechain (approved):

api/admin/health/health.go
```diff
@@ -0,0 +1,110 @@
...
+
+type BlockIngestion struct {
+	ID        *thor.Bytes32 `json:"id"`
+	Timestamp *time.Time    `json:"timestamp"`
+}
+
+type Status struct {
+	Healthy           bool            `json:"healthy"`
+	BlockIngestion    *BlockIngestion `json:"blockIngestion"`
+	ChainBootstrapped bool            `json:"chainBootstrapped"`
```
libotony, line 25: `ChainBootstrapped` can be removed considering `BlockIngestion`?

darrenvechain (dismissed):

api/admin/health/health.go
```diff
@@ -0,0 +1,106 @@
...
+)
+
+type BlockIngestion struct {
+	ID        *thor.Bytes32 `json:"id"`
+	Timestamp *time.Time    `json:"timestamp"`
+}
+
+type Status struct {
+	Healthy            bool       `json:"healthy"`
+	BestBlockTimestamp *time.Time `json:"bestBlockTimestamp"`
```
libotony, line 24: What value will this be encoded to JSON? String or number?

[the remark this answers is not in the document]
otherview: time.time JSON serialization uses a well known standard, ISO 8601 for JSON. We can use uint64, no strong feelings about it.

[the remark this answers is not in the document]
otherview: I'm starting to think that human-readable might be a good trait to have, as it helps in readability.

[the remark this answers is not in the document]
libotony: If you consider readability is first priority, how about updating the naming to `BestBlockTime`.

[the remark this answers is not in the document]
otherview: Yeah, makes sense 👍

cmd/thor/main.go
```diff
@@ -310,6 +316,9 @@ func soloAction(ctx *cli.Context) error {
 
 	logLevel := initLogger(lvl, ctx.Bool(jsonLogsFlag.Name))
 
+	onDemandBlockProduction := ctx.Bool(onDemandFlag.Name)
+	blockProductionInterval := ctx.Uint64(blockInterval.Name)
```
libotony, line 320: Check of interval is missing
```
if blockInterval == 0 {
    return errors.New("block-interval cannot be zero")
}
```

libotony (dismissed):

cmd/thor/main.go
otherview:
```suggestion
	blockProductionInterval := ctx.Uint64(blockInterval.Name)
	if blockInterval == 0 {
		return errors.New("block-interval cannot be zero")
	}

```

paologalligit (approved):

PR #845, opened 2024-09-12
pool: track pending cost per payer

libotony:
# Description

Addresses vechain/protocol-board-repo#311

# Checklist:

### DIFF
--- a/runtime/resolved_tx.go
+++ b/runtime/resolved_tx.go
@@ -114,6 +114,7 @@ func (r *ResolvedTransaction) BuyGas(state *state.State, blockTime uint64) (
 		return returnedEnergy, nil
 	}
 
+	// prepaid is the max total of gas cost available to spend on this transaction
 	prepaid = new(big.Int).Mul(new(big.Int).SetUint64(r.tx.Gas()), gasPrice)
 	if r.Delegator != nil {
 		var sufficient bool
--- a/runtime/resolved_tx.go
+++ b/runtime/resolved_tx.go
@@ -96,6 +96,7 @@ func (r *ResolvedTransaction) BuyGas(state *state.State, blockTime uint64) (
 	baseGasPrice *big.Int,
 	gasPrice *big.Int,
 	payer thor.Address,
+	prepaid *big.Int,
 	returnGas func(uint64) error,
 	err error,
 ) {
@@ -113,19 +114,19 @@ func (r *ResolvedTransaction) BuyGas(state *state.State, blockTime uint64) (
 		return returnedEnergy, nil
 	}
 
-	prepaid := new(big.Int).Mul(new(big.Int).SetUint64(r.tx.Gas()), gasPrice)
+	prepaid = new(big.Int).Mul(new(big.Int).SetUint64(r.tx.Gas()), gasPrice)
 	if r.Delegator != nil {
 		var sufficient bool
 		if sufficient, err = energy.Sub(*r.Delegator, prepaid); err != nil {
 			return
 		}
 		if sufficient {
-			return baseGasPrice, gasPrice, *r.Delegator, func(rgas uint64) error {
+			return baseGasPrice, gasPrice, *r.Delegator, prepaid, func(rgas uint64) error {
 				_, err := doReturnGas(rgas)
 				return err
 			}, nil
 		}
-		return nil, nil, thor.Address{}, nil, errors.New("insufficient energy")
+		return nil, nil, thor.Address{}, nil, nil, errors.New("insufficient energy")
 	}
 
 	commonTo := r.CommonTo()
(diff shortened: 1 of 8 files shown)


txpool/tx_object_map.go
```diff
@@ -45,26 +51,53 @@ func (m *txObjectMap) Add(txObj *txObject, limitPerAccount int) error {
...
-		m.quota[*d]++
 	}
 
-	m.quota[txObj.Origin()]++
+	var (
+		cost  *big.Int
+		payer thor.Address
+	)
+
+	if txObj.Cost() != nil {
```
otherview, line 71: Is it possible to flood the mempool with no-cost transactions ?

txpool/tx_object_map.go
```diff
@@ -73,17 +106,29 @@ func (m *txObjectMap) RemoveByHash(txHash thor.Bytes32) bool {
...
+			delegator := delegatorKey(*d)
+			if m.quota[delegator] > 1 {
+				m.quota[delegator]--
 			} else {
-				delete(m.quota, *d)
+				delete(m.quota, delegator)
+			}
+		}
+
+		if payer := txObj.Payer(); payer != nil {
```
otherview, line 125:
```suggestion
		// update the pending list and current map of payers
		if payer := txObj.Payer(); payer != nil {
```

txpool/tx_object_map.go
```diff
@@ -94,6 +139,17 @@ func (m *txObjectMap) RemoveByHash(txHash thor.Bytes32) bool {
 	return false
 }
 
+func (m *txObjectMap) UpdateCost(txObj *txObject) {
```
otherview, line 142: Minor sugestion:
```suggestion
func (m *txObjectMap) UpdatePendingCost(txObj *txObject) {
```

txpool/tx_object_map.go
```diff
@@ -45,26 +51,53 @@ func (m *txObjectMap) Add(txObj *txObject, limitPerAccount int) error {
...
+
+	m.quota[origin]++
 	m.mapByHash[hash] = txObj
 	m.mapByID[txObj.ID()] = txObj
+	if delegator != nil {
+		m.quota[*delegator]++
+	}
+	if cost != nil {
+		m.cost[payer] = cost
+	}
```
otherview, line 94: Minor suggestion :

```suggestion
	m.quota[origin]++
	if delegator != nil {
		m.quota[*delegator]++
	}
	
	m.mapByHash[hash] = txObj
	m.mapByID[txObj.ID()] = txObj
	
	if cost != nil {
		m.cost[payer] = cost
	}
```

otherview (commented):
Good stuff. It looks good to me, only that concern around possible no-cost txs ?

txpool/tx_object_map.go
```diff
@@ -45,26 +51,53 @@ func (m *txObjectMap) Add(txObj *txObject, limitPerAccount int) error {
 		return nil
 	}
 
-	if m.quota[txObj.Origin()] >= limitPerAccount {
+	origin := originKey(txObj.Origin())
```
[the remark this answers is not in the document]
libotony: I think the last three questions are related.
> Are these types used ?
> What if quota map[interface{}]int was quota map[thor.Address]int ?
> What if there was no originKey cast here ?

The two new types are designed to handle the situation for one account both being transaction origin and transaction delegator.

txpool/tx_object_map.go
libotony: Not possible at peer-to-peer level, possible at API level, won't propagate until executable.

runtime/resolved_tx.go
```diff
@@ -113,19 +114,20 @@ func (r *ResolvedTransaction) BuyGas(state *state.State, blockTime uint64) (
...
 	}
 
-	prepaid := new(big.Int).Mul(new(big.Int).SetUint64(r.tx.Gas()), gasPrice)
+	// prepaid is the max total of gas cost available to spend on this transaction
+	prepaid = new(big.Int).Mul(new(big.Int).SetUint64(r.tx.Gas()), gasPrice)
 	if r.Delegator != nil {
 		var sufficient bool
 		if sufficient, err = energy.Sub(*r.Delegator, prepaid); err != nil {
 			return
 		}
```
otherview, line 123: What do you think about this empty returns ? 👀

txpool/tx_object_map.go
otherview: It seems like they wouldn't be added until executable right ? So we should be good.

However theres seems to be a different behaviour [here](https://github.com/vechain/thor/blob/cc2b3764839ce910d979cc6d1c283d60acfc17e6/txpool/tx_pool.go#L279-L294). Let's pair up and review it separately to this PR tomorrow perhaps ?

otherview (approved):

PR #846, opened 2024-09-12
Add new gosec gh yaml file

paologalligit:
# Description

Add new workflow for gosec when creating PRs to master and when merging into master.

[GH Issue](https://github.com/vechain/protocol-board-repo/issues/310)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing E2E tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have not added any vulnerable dependencies to my code

PR #852, opened 2024-09-23
switch to math/rand/v2

libotony:
# Description

This PR switched to math/rand/v2, which enhanced the randomness along with improved API design. Also moved the gosec action into lint.yaml.

# Checklist:


.github/workflows/gosec.yaml
```diff
@@ -1,25 +0,0 @@
-name: Gosec
```
otherview, line 1: I think that perhaps this file should have more comments explaining the purpose behind this GH action.

The idea was to have a cutting-edge go-sec check that runs against go-sec master branch while allowing to (at times ) be unstable. So even if it fails it should not meant a stop-of-the-world for the CD/CI.

Perhaps keeping this pipeline feature isolated from other stabler linting features could be a good idea.

.golangci.yml
```diff
@@ -24,7 +24,6 @@ linters:
     - durationcheck
     - copyloopvar
     - whitespace
-    - gosec
```
otherview, line 27: I think we still want gosec to run on the regular linter.

libotony:
Consensus from after standup discussion:

The linting process should maintain the same behavior as the CI/CD pipeline, so we have two options:

+ Keep `gosec` in `golanglint-ci` and make standalone `gosec` action only reports warning
Or
+ If step one is not possible, add `gosec` into `make lint`

@otherview @paologalligit

darrenvechain:
@libotony shouldn't gosec be fully configurable inside golanglint?

https://golangci-lint.run/usage/linters/#gosec

libotony:
> @libotony shouldn't gosec be fully configurable inside golanglint?
> 
> https://golangci-lint.run/usage/linters/#gosec

#846 Added the standalone gosec action, which is updated to the latest gosec while the gosec inside `golanglint-ci` is falling behind.

darrenvechain:
@libotony looks like we just need to update to latest golanglint:

https://github.com/golangci/golangci-lint/blob/v1.61.0/go.mod#L95

block/header_test.go
```diff
@@ -6,7 +6,7 @@
 package block
 
 import (
-	"math/rand"
+	"crypto/rand"
```
leszek-vechain, line 9: should it also be math/v2 ?

libotony: Nope, `math/rand/v2` does not provide `rand.Read` function

darrenvechain (approved):

otherview:
> Consensus from after standup discussion:
> 
> The linting process should maintain the same behavior as the CI/CD pipeline, so we have two options:
> 
> * Keep `gosec` in `golanglint-ci` and make standalone `gosec` action only reports warning
>   Or
> * If step one is not possible, add `gosec` into `make lint`
> 
> @otherview @paologalligit

I think this got a bit muddled over the last few weeks.

AFAIR the decision was to maintain the current linter while updating it as need AND introduce a new GH action specific for gosec running at the latest version.

otherview (approved):

otherview (approved):

darrenvechain (approved):

PR #861, opened 2024-10-10
Darren/feat/add subscription cache

darrenvechain:
# Description

This PR leverages an LRU cache for `beat`, `beat2` and `block` subscription endpoints. It generates the messages, marshals them and caches the bytes against thor block ID.

This improves CPU performance a lot when:
- There are a high amount of transactions, eg 90% gas utilization
- There are high number of beat/ beat2 websocket connections

# How Has This Been Tested?

- [x] Performance tested using https://github.com/darrenvechain/toolchain & a simple k6 script to create 10,000 websockets with an isolated node on testnet.
- [ ] Unit tests

# Checklist:

### DIFF
--- a/api/subscriptions/subscriptions.go
+++ b/api/subscriptions/subscriptions.go
@@ -31,10 +31,13 @@ type Subscriptions struct {
 	pendingTx      *pendingTx
 	done           chan struct{}
 	wg             sync.WaitGroup
+	beat2Cache     *messageCache
+	beatCache      *messageCache
+	blockCache     *messageCache
 }
 
 type msgReader interface {
-	Read() (msgs []interface{}, hasMore bool, err error)
+	Read() (msgs [][]byte, hasMore bool, err error)
 }
 
 var (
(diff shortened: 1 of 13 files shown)


api/subscriptions/subscriptions.go
```diff
@@ -49,6 +52,11 @@ const (
 )
 
 func New(repo *chain.Repository, allowedOrigins []string, backtraceLimit uint32, txpool *txpool.TxPool) *Subscriptions {
+	cacheSize := backtraceLimit * 120 / 100
+	beat2Cache, _ := newMessageCache(generateBeat2Message, int(cacheSize))
+	beatCache, _ := newMessageCache(generateBeatMessage, int(cacheSize))
+	blockCache, _ := newMessageCache(generateBlockMessage, int(cacheSize))
```
darrenvechain, line 58: Ignoring errors for now, as this will only error if the cache size is less than 0. Should we handle this better?

`subscriptions.New` and `api.New` do not currently return errors

api/subscriptions/subscriptions.go
```diff
@@ -49,6 +52,11 @@ const (
 )
 
 func New(repo *chain.Repository, allowedOrigins []string, backtraceLimit uint32, txpool *txpool.TxPool) *Subscriptions {
+	cacheSize := backtraceLimit * 120 / 100
```
darrenvechain, line 55: This may be problematic if node hosters set a very large number for the backtrace limit.

The idea around the `120` is to add %20 extra capacity to the backtrace limit.

Should we set a constant max instead of using the backtrace?

api/subscriptions/subscriptions.go
leszek-vechain: I guess we can check if cache is > 0 what is the point of having cache < 0 ?

darrenvechain: the `backtraceLimit` is already uint32, so it can't be negative

leszek-vechain: I see, maybe just prevent from being 0 then as it makes no sense for 0 cache ?

PR #863, opened 2024-10-15
Add additional block tests

MakisChristou:


consensus/consensus_test.go
```diff
@@ -650,6 +699,28 @@ func TestValidateBlockBody(t *testing.T) {
 				assert.Equal(t, err, expected)
 			},
 		},
+		{
+			"EmptyTx", func(t *testing.T) {
```
paologalligit, line 703: Maybe this can be renamed to `ZeroGasTx` since it's focusing on sending a tx with 0 gas

leszek-vechain (approved):

PR #865, opened 2024-10-21
Darren/logdb remove leading zeros

darrenvechain:
# Description

This changes reduces the storage size for the LogDB by removing leading zeros from topics. Eg, instead of storing topic 1 equal to:

```
0000000000000000000000006d95e6dca01d109882fe1726a2fb9865fa41e7aa
```

We store:
```
6d95e6dca01d109882fe1726a2fb9865fa41e7aa
```

This saves over 0.5 GB of storage.

## Type of change

- [x] Breaking change - requires a full resync of the logDB

# How Has This Been Tested?

- [x] Unit tested
- [x] [Performance tested](https://github.com/darrenvechain/thor-performance-testing) - No change in query performance
- [x] [Comparison tested](https://github.com/darrenvechain/thor-comparative-testing/blob/main/tests/endpoints/logs/events/compare-events.test.ts) - Mainnet against synced node with these changes. All API calls returned the same values for the same request

# Checklist:


logdb/logdb.go
```diff
@@ -398,11 +398,24 @@ func (db *LogDB) NewWriterSyncOff() *Writer {
...
 	return nil
 }
 
+func removeLeadingZeros(bytes []byte) []byte {
+	i := 0
+	// increase i until it reaches the first non-zero byte
+	for ; i < len(bytes) && bytes[i] == 0; i++ {
+	}
+	bytes = bytes[i:]
+	// ensure at lease 1 byte
```
paologalligit, line 412:
```suggestion
	// ensure at least 1 byte
```

paologalligit (approved):
Looks good to me! Just a minor typo

logdb/logdb.go
```diff
@@ -398,11 +398,24 @@ func (db *LogDB) NewWriterSyncOff() *Writer {
...
 }
 
+func removeLeadingZeros(bytes []byte) []byte {
+	i := 0
+	// increase i until it reaches the first non-zero byte
+	for ; i < len(bytes) && bytes[i] == 0; i++ {
+	}
+	bytes = bytes[i:]
+	// ensure at least 1 byte
+	if len(bytes) == 0 {
```
libotony, line 413: Any particular reason for this?

logdb/logdb.go
```diff
@@ -474,14 +487,20 @@ func (w *Writer) Write(b *block.Block, receipts tx.Receipts) error {
 
 		for clauseIndex, output := range r.Outputs {
 			for _, ev := range output.Events {
+				t0Val := topicValue(ev.Topics, 0)
```
libotony, line 490: Looks the same as changing before.

libotony: NVM, I got it.

logdb/logdb.go
libotony: NVM, I got that.

logdb/logdb.go
darrenvechain: Apologies, this was me debugging I think, I'll revert

qianbin:
It seems not very useful on reducing the size of logdb, And it's not worth it.

logdb/logdb.go
darrenvechain: Pushed this change a change for:
```golang
if i == len(bytes) {
	return []byte{0}
}
```
because the unit test was failing when you pass in a number of bytes not equal to 32

darrenvechain:
> It seems not very useful on reducing the size of logdb, And it's not worth it.

@qianbin the idea is to put it with mainDB v4 changes, as a full sync is required. There's not really any negatives to include it there.

qianbin:
> > It seems not very useful on reducing the size of logdb, And it's not worth it.
> 
> @qianbin the idea is to put it with mainDB v4 changes, as a full sync is required. There's not really any negatives to include it there.

It'll be good if so.

However, I think the compressed(zstd ? gzip?) VFS of sqlite might be the more advanced solution.

PR #866, opened 2024-10-21
Darren/feat/add subscription cache

darrenvechain:
# Description

This PR leverages a simple cache for `beat`, `beat2` and `block` subscription endpoints. It generates the messages,  caches the struct against thor block ID.

This improves CPU performance a lot when:
- There are a high amount of transactions, eg 90% gas utilization
- There are high number of beat/ beat2 websocket connections

# How Has This Been Tested?

- [x] Performance tested using https://github.com/darrenvechain/toolchain & a simple k6 script to create 10,000 websockets with an isolated node on testnet.

# Checklist:

### DIFF
--- a/.github/workflows/on-pull-request.yaml
+++ b/.github/workflows/on-pull-request.yaml
@@ -3,7 +3,7 @@ name: Pull Request CI
 on:
   pull_request:
     branches:
-      - master
+      - '*'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
--- a/api/subscriptions/message_cache.go
+++ b/api/subscriptions/message_cache.go
@@ -1,3 +1,8 @@
+// Copyright (c) 2024 The VeChainThor developers
+//
+// Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
+// file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
+
 package subscriptions
 
 import (
--- a/api/subscriptions/message_cache.go
+++ b/api/subscriptions/message_cache.go
@@ -6,32 +6,39 @@
 package subscriptions
 
 import (
+	"fmt"
 	"sync"
 
 	lru "github.com/hashicorp/golang-lru"
-	"github.com/vechain/thor/v2/chain"
+	"github.com/vechain/thor/v2/thor"
 )
 
-type messageHandler = func(*chain.ExtendedBlock, *chain.Repository) ([]byte, error)
-
 type messageCache struct {
-	cache     *lru.Cache
-	generator messageHandler
-	mu        sync.RWMutex
+	cache *lru.Cache
+	mu    sync.RWMutex
 }
 
-func newMessageCache(handler messageHandler, cacheSize int) (*messageCache, error) {
-	cache, err := lru.New(cacheSize)
+func newMessageCache(cacheSize uint32) *messageCache {
+	if cacheSize > 1000 {
+		cacheSize = 1000
+	}
+	if cacheSize == 0 {
+		cacheSize = 1
+	}
+	cache, err := lru.New(int(cacheSize))
+	if err != nil {
+		// lru.New only throws an error if the number is less than 1
+		panic(fmt.Errorf("failed to create message cache: %v", err))
+	}
 	return &messageCache{
-		cache:     cache,
-		generator: handler,
-	}, err
+		cache: cache,
+	}
 }
 
 // GetOrAdd returns the message of the block, if the message is not in the cache, it will generate the message and add it to the cache.
 // The second return value indicates whether the message is newly generated.
-func (mc *messageCache) GetOrAdd(block *chain.ExtendedBlock, repo *chain.Repository) ([]byte, bool, error) {
-	blockID := block.Header().ID().String()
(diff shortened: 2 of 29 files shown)


api/subscriptions/message_cache.go
```diff
@@ -0,0 +1,62 @@
...
+
+func newMessageCache(cacheSize uint32) *messageCache {
+	if cacheSize > 1000 {
+		cacheSize = 1000
+	}
+	if cacheSize == 0 {
+		cacheSize = 1
+	}
+	cache, err := lru.New(int(cacheSize))
+	if err != nil {
```
darrenvechain, line 29: TODO: Can we handle this more gracefully? Currently `subscriptions.New` nor `api.New` don't return errors, so this would mess up the flow a little bit

api/subscriptions/subscriptions.go
```diff
@@ -308,7 +314,7 @@ func (s *Subscriptions) pipe(conn *websocket.Conn, reader msgReader, closed chan
 			return err
 		}
 		for _, msg := range msgs {
-			if err := conn.WriteJSON(msg); err != nil {
+			if err := conn.WriteMessage(websocket.BinaryMessage, msg); err != nil {
```
libotony, line 317: Message type should be `TextMessage` ?

api/subscriptions/subscriptions.go
```diff
@@ -67,8 +70,11 @@ func New(repo *chain.Repository, allowedOrigins []string, backtraceLimit uint32,
 				return false
 			},
 		},
-		pendingTx: newPendingTx(txpool),
-		done:      make(chan struct{}),
+		pendingTx:  newPendingTx(txpool),
+		done:       make(chan struct{}),
+		beat2Cache: newMessageCache(backtraceLimit),
+		beatCache:  newMessageCache(backtraceLimit),
+		blockCache: newMessageCache(backtraceLimit),
```
libotony, line 77: Block cache is not necessary since there is no computing.

api/subscriptions/message_cache.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/vechain/thor/v2/thor"
+)
+
+type messageCache struct {
```
libotony, line 16: Seems not necessarily creating a new type here, make it in-place within xxx_reader sounds better to me. wdyt?

api/subscriptions/message_cache.go
```diff
@@ -0,0 +1,62 @@
...
+	return &messageCache{
+		cache: cache,
+	}
+}
+
+// GetOrAdd returns the message of the block, if the message is not in the cache, it will generate the message and add it to the cache.
+// The second return value indicates whether the message is newly generated.
+func (mc *messageCache) GetOrAdd(id thor.Bytes32, createMessage func() ([]byte, error)) ([]byte, bool, error) {
+	blockID := id.String()
+	mc.mu.RLock()
```
libotony, line 42: Lru cache has a built-in mutex, you might need the `simple-lru` to implement your own locked version.

libotony: Or no own managed lock fine seems fit.

api/subscriptions/message_cache.go
darrenvechain: We can do that for sure, but the `newXXXReader` gets initialised for each new subscription, so we'd have to think about how we handle the cache outside of the individual subs

libotony: You are right, I have missed that part.

api/subscriptions/message_cache.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru"
+	"github.com/vechain/thor/v2/thor"
+)
+
+// messageCache is a generic cache that stores messages of any type.
+type messageCache[T any] struct {
+	cache *lru.Cache
```
libotony, line 18: In this scenario, should use lock free versioned `github.com/hashicorp/golang-lru/simplelru`

api/subscriptions/message_cache.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+}
+
+// GetOrAdd returns the message of the block. If the message is not in the cache,
+// it will generate the message and add it to the cache. The second return value
+// indicates whether the message is newly generated.
+func (mc *messageCache[T]) GetOrAdd(id thor.Bytes32, createMessage func() (T, error)) (T, bool, error) {
+	blockID := id.String()
+	mc.mu.RLock()
+	msg, ok := mc.cache.Get(blockID)
```
libotony, line 46:
```suggestion
	msg, ok := mc.cache.Get(id)
```

`thor.Bytes32` should be fine here.

otherview (approved):
Minor tweak - lgtm otherwise

.github/workflows/on-pull-request.yaml
```diff
@@ -3,7 +3,7 @@ name: Pull Request CI
 on:
   pull_request:
     branches:
-      - master
+      - '*'
```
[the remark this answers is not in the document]
darrenvechain: Yepp, because PRs don't run CI if creating PRs into other branches, which is annoying for failures

PR #869, opened 2024-10-29
Show all issues on lint

otherview:
# Description

It looks like the thor client merge got away with passing some linter issues to the master code.
This mainly happened because the `only-new-issues` was active.
The fix is to guarantee the `only-new-issues` is deactivated.

Fixes # [(issue)](https://github.com/vechain/otherviews-workboard/issues/106)

# Checklist:

### DIFF
--- a/.github/workflows/lint-go.yaml
+++ b/.github/workflows/lint-go.yaml
@@ -23,5 +23,5 @@ jobs:
           version: v1.60.3
           # use the default if on main branch, otherwise use the pull request config
           args: --timeout=30m --config=.golangci.yml
-          only-new-issues: true
+          only-new-issues: false
           skip-cache: true
--- a/api/accounts/accounts_test.go
+++ b/api/accounts/accounts_test.go
@@ -6,10 +6,8 @@
 package accounts_test
 
 import (
-	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"math/big"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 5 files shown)


.github/workflows/lint-go.yaml
```diff
@@ -23,5 +23,5 @@ jobs:
           version: v1.60.3
           # use the default if on main branch, otherwise use the pull request config
           args: --timeout=30m --config=.golangci.yml
-          only-new-issues: true
+          only-new-issues: false
```
darrenvechain, line 26:
```suggestion
          only-new-issues: ${{ github.event_name == 'pull_request' }}
```

api/accounts/accounts_test.go
```diff
@@ -578,45 +576,3 @@ func batchCallWithNonExistingRevision(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, statusCode, "bad revision")
 	assert.Equal(t, "revision: leveldb: not found\n", string(res), "revision not found")
 }
-
-func httpPost(t *testing.T, url string, body interface{}) ([]byte, int) {
```
leszek-vechain, line 582: why we remove all these methods ? are they somewhere else or we don't use them at all ?

.github/workflows/lint-go.yaml
otherview: This seems to assume that it's expected only to detect new issues in pull requests, which I believe may fail in certain edge cases like the scenario that brought us here.
While there's a benefit of only running the lint for the diffs, golangci-lint linter is so fast that the difference looks negligible ?

api/accounts/accounts_test.go
otherview: They're not used any longer. I think they were made redundant by the thor client merge.

darrenvechain (approved):

PR #873, opened 2024-11-05
enhancement(logging): leverage trace level

darrenvechain:
# Description

Leverage the trace log level so we can enable `Debug` without getting spammed with logs.

# Testing 

```bash
make thor
./bin/thor --network main --verbosity 4
```

Restarting a node that is nearly synced before starting:

image

Starting a node with a lot of blocks to sync:

image

# Checklist:

### DIFF
--- a/cmd/thor/node/node.go
+++ b/cmd/thor/node/node.go
@@ -266,7 +266,7 @@ func (n *Node) txStashLoop(ctx context.Context) {
 			if err := stash.Save(txEv.Tx); err != nil {
 				logger.Warn("stash tx", "id", txEv.Tx.ID(), "err", err)
 			} else {
-				logger.Debug("stashed tx", "id", txEv.Tx.ID())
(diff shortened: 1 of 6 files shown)


otherview (approved):
lgtm - optionally there are some logs that could benefit with ids, or msgTypes
But given that we're making them less visible, not sure it's worth adding that info ?

libotony (approved):
Nice catch！

darrenvechain:
> looks good, any other places worth switching ? 

I think its fine for now, log levels are greatly reduced now, and its usually just a debug when a peer connects.

darrenvechain:
> lgtm - optionally there are some logs that could benefit with ids, or msgTypes But given that we're making them less visible, not sure it's worth adding that info ?

Think we already have this, eg.

```golang
log := peer.logger.New("msg", proto.MsgName(msg.Code))
log.Trace("received RPC call")
```

PR #881, opened 2024-11-18
Update Convert Filter to match BlockMask bit space

otherview:
# Description

Modifies the `ConvertFilter` method to take into account that the max block byte space is of 28 bytes instead of 32.

Now returns error if a direct access to the DB requests more than `blockNumMask`

### DIFF
--- a/api/events/types.go
+++ b/api/events/types.go
@@ -173,8 +173,21 @@ func ConvertRange(chain *chain.Chain, r *Range) (*logdb.Range, error) {
 			To:   toHeader.Number(),
 		}, nil
 	}
+
+	// Units are blocks, locked a max of 28bits in the logdb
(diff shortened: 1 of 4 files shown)


api/events/types.go
```diff
@@ -173,8 +173,21 @@ func ConvertRange(chain *chain.Chain, r *Range) (*logdb.Range, error) {
 			To:   toHeader.Number(),
 		}, nil
 	}
+
+	// Units are blocks, locked a max of 28bits in the logdb
+	maxRange := uint32(1<<28 - 1) // Maximum value for 28 bits: 2^28 - 1
```
paologalligit, line 178: Can we take this value from the constant in sequence.go? So we don't have to deal with scattered values

otherview: Good point ! Fixed.

darrenvechain (approved):

PR #882, opened 2024-11-18
Replace cache.ShouldLog with Log

otherview:
# Description

It's a simplification of the `muxdb/cache.go` Log to always log the cache stats instead of delaying them.
`cacheStats` logging is now controlled by the cache and always printed in 20sec interval unless unchanged.

### DIFF
--- a/muxdb/cache.go
+++ b/muxdb/cache.go
@@ -51,13 +51,8 @@ func (c *cache) log() {
 	last := c.lastLogTime.Swap(now)
 
 	if now-last > int64(time.Second*20) {
-		log1, ok1 := c.nodeStats.ShouldLog("node cache stats")
-		log2, ok2 := c.rootStats.ShouldLog("root cache stats")
-
-		if ok1 || ok2 {
-			log1()
-			log2()
-		}
+		c.nodeStats.Log("node cache stats")
+		c.rootStats.Log("root cache stats")
 	} else {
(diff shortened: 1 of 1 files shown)


muxdb/cache.go
```diff
@@ -51,13 +51,8 @@ func (c *cache) log() {
...
 
 	if now-last > int64(time.Second*20) {
-		log1, ok1 := c.nodeStats.ShouldLog("node cache stats")
-		log2, ok2 := c.rootStats.ShouldLog("root cache stats")
-
-		if ok1 || ok2 {
-			log1()
-			log2()
-		}
+		c.nodeStats.Log("node cache stats")
```
freemanzMrojo, line 54: The overarching method is also called `log`. Even though it applies to `cache` instead, should we rename any of them?

I also think that the former `ShouldLog` applied just within the package, should it be lower-case (not sure if we are being strict about this though, just my first review, feel free to dismiss this :p )

freemanzMrojo (approved):
LGTM, small comment since I was working on the same files

libotony:
`ShouldLog` has an improvement to prevent a lot of log after node is synced.

libotony:
Closing this PR in favor of thie commit https://github.com/vechain/thor/pull/868/commits/26f1514a2117ec548c55b7b33f2d1dc0cf7c568a

PR #883, opened 2024-11-18
Add dummy cache for inmem ops

otherview:
# Description

This PR adds an Empty Cache structure that can be used when using inmem Mux DBs.

In essence, this removes the `if c == nil` statement where `c` is the receiver, fix that was created to handle inmem dbs.

No strong feelings towards merging this, but decided to raise the PR as it felt weird when a method starts by testing if the instance is nil.


muxdb/empty_cache.go
```diff
@@ -0,0 +1,22 @@
+package muxdb
+
+import "github.com/vechain/thor/v2/trie"
+
+type emptyCache struct{}
```
libotony, line 5: How about `dummyCache` and put into `cache.go`?

otherview: Good idea 👍  Fixed!

libotony (approved):

PR #885, opened 2024-11-19
return error in newSequence

libotony:
# Description

Return error in `newSequence` to avoid unexpected process exit.

# Checklist:

### DIFF
--- a/api/events/types.go
+++ b/api/events/types.go
@@ -174,19 +174,18 @@ func ConvertRange(chain *chain.Chain, r *Range) (*logdb.Range, error) {
 		}, nil
 	}
 
-	// Units are blocks, locked a max of 28bits in the logdb
-	from := uint32(r.From)
-	to := uint32(r.To)
+	rng := logdb.Range{
+		From: uint32(r.From),
+		To:   uint32(r.To),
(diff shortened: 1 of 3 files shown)


api/events/types.go
```diff
@@ -174,19 +174,18 @@ func ConvertRange(chain *chain.Chain, r *Range) (*logdb.Range, error) {
...
+	rng := logdb.Range{
+		From: uint32(r.From),
+		To:   uint32(r.To),
+	}
 
-	// Ensure the values are capped at the 28-bit maximum
-	if uint32(r.From) > logdb.BlockNumMask {
-		from = logdb.BlockNumMask
+	headNum := block.Number(chain.HeadID())
+	if rng.From > headNum {
```
darrenvechain, line 183: If client passes a `from` larger than the best block, they might receive events even though they don't expect any. Might be a bit of a weird use case though

darrenvechain (approved):

api/events/types.go
libotony: You're right, here is just a safety guard.

paologalligit (approved):

PR #886, opened 2024-11-19
Metrics: Cache hit/miss

freemanzMrojo:
# Description

Covers the metrics for cache hit/miss for root and node cache. In the diagram below:

1. Yellow line is a hit for the root one (in-memory representation of the trie)
2. Green line is a hit for a node one
3. Blue one is a miss for a node one
4. Red one is a miss for the root one

The formula in Grafana is `rate(thor_metrics_cache_hit_miss_count[$__rate_interval])`

Screenshot 2024-11-19 at 10 42 57

### DIFF
--- a/muxdb/cache.go
+++ b/muxdb/cache.go
@@ -179,12 +179,8 @@ type cacheStats struct {
 	flag      atomic.Int32
 }
 
-func (cs *cacheStats) Hit() int64  {
-	return cs.hit.Add(1) 
-}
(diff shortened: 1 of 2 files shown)


darrenvechain (approved):

libotony:
@freemanzMrojo It seems to be an expensive operation, could you give it another consideration? Considering trie read/write is the most frequent operation in the node. Could you give it another consideration? Maybe follow the logic of `ShouldLog`?

PR #888, opened 2024-11-20
Metrics: Reuse `shouldLog` so we get cache hit/miss data at the same pace

freemanzMrojo:
# Description

After a conversation with @libotony in a previous PR, creating this one so we get the metrics data at the same pace as we log the hit rate.

1. Raw hit/miss by number

Screenshot 2024-11-20 at 14 43 26

- Yellow line: Root node hit rate
- Green line: Other trie node hit rate
- Orange line: Root node miss rate
- Blue line: Other trie node miss rate

2. Rate in percentage (hits over lookups)

Screenshot 2024-11-20 at 15 01 48

- Yellow line: Root trie node
- Green line: Other trie node

### DIFF
--- a/muxdb/cache.go
+++ b/muxdb/cache.go
@@ -109,7 +109,6 @@ func (c *cache) GetNodeBlob(keyBuf *[]byte, name string, path []byte, ver trie.V
 	}, peek) && len(blob) > 0 {
 		if !peek {
 			c.nodeStats.Hit()
-			metricCacheHitMissCounterVec().AddWithLabel(1, map[string]string{"type": "node", "event": "hit"})
 		}
 		return blob
 	}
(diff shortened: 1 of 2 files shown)


darrenvechain:
@freemanzMrojo graph looks a bit hard to digest, could you try something like this:

```
100 * (
    sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="hit"}[5m])) /
    (
        sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="hit"}[5m])) +
        sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="miss"}[5m]))
    )
)
```

freemanzMrojo:
> @freemanzMrojo graph looks a bit hard to digest, could you try something like this:
> 
> ```
> 100 * (
>     sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="hit"}[5m])) /
>     (
>         sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="hit"}[5m])) +
>         sum(rate(thor_metrics_cache_hit_miss_count{type="root", event="miss"}[5m]))
>     )
> )
> ```
> 
> I'm curios how it will look as time goes on
> 
> PS: maybe use the __rate_interval var from grafana or whatever its called, instead of `5m`

Updated the PR description with this @darrenvechain

darrenvechain (approved):

PR #891, opened 2024-11-22
Pedro/maindb v4/benchmarks

otherview:
# Description

Benchmark tests for Node Block Process


cmd/thor/node/node_benchmark_test.go
```diff
@@ -0,0 +1,374 @@
...
+)
+
+func getCachedAccounts(b *testing.B) []genesis.DevAccount {
+	once.Do(func() {
+		cachedAccounts = createAccounts(b, 1_000)
+	})
+	return cachedAccounts
+}
+
+func BenchmarkBlockProcess_RandomSigners_ManyClausesPerTx(b *testing.B) {
```
darrenvechain, line 41: are these the same tests as the ones in the #892  ?

It might be better to merge that to master, then we can sync these branches with master

darrenvechain: I'd suggest we merge the other PR then and close this one. We can sync mainDB branch with master once the other is merged. Otherwise we get a merge nightmare

otherview: Yes fair concern, I've been keeping tabs on the differences between them, so we should be good to go 👍

PR #893, opened 2024-11-25
chore: update API metrics bucket and endpoint names

darrenvechain:
# Description

- Make endpoint names more readable in the metrics dashboards
- Update the bucket. Previously we weren't capturing meaningful buckets because 99% of requests were between 0-150ms

# Checklist:

### DIFF
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -64,7 +64,7 @@ func metricsMiddleware(next http.Handler) http.Handler {
 			name         = ""
 			subscription = ""
 		)
-
+		
 		// all named route will be recorded
 		if rt != nil && rt.GetName() != "" {
 			enabled = true
--- a/api/transactions/transactions.go
+++ b/api/transactions/transactions.go
@@ -218,14 +218,14 @@ func (t *Transactions) Mount(root *mux.Router, pathPrefix string) {
 
 	sub.Path("").
 		Methods(http.MethodPost).
-		Name("transactions_send_tx").
+		Name("POST /transactions").
 		HandlerFunc(utils.WrapHandlerFunc(t.handleSendTransaction))
 	sub.Path("/{id}").
 		Methods(http.MethodGet).
-		Name("transactions_get_tx").
+		Name("GET /transactions/${id}").
 		HandlerFunc(utils.WrapHandlerFunc(t.handleGetTransactionByID))
 	sub.Path("/{id}/receipt").
 		Methods(http.MethodGet).
-		Name("transactions_get_receipt").
+		Name("GET /transactions/${id}/receipt").
 		HandlerFunc(utils.WrapHandlerFunc(t.handleGetTransactionReceiptByID))
 }
--- a/metrics/telemetry.go
+++ b/metrics/telemetry.go
@@ -30,7 +30,11 @@ func HTTPHandler() http.Handler {
 // Define standard buckets for histograms
 var (
 	Bucket10s      = []int64{0, 500, 1000, 2000, 3000, 4000, 5000, 7500, 10_000}
-	BucketHTTPReqs = []int64{0, 150, 300, 450, 600, 900, 1200, 1500, 3000}
+	BucketHTTPReqs = []int64{
+		0, 1, 2, 5, 10, 20, 30, 50, 75, 100,
+		150, 200, 300, 400, 500, 750, 1000,
+		1500, 2000, 3000, 4000, 5000, 10000,
+	}
 )
 
 // HistogramMeter represents the type of metric that is calculated by aggregating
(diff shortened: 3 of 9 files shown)


leszek-vechain (approved):

api/transactions/transactions.go
```diff
@@ -218,14 +218,14 @@ func (t *Transactions) Mount(root *mux.Router, pathPrefix string) {
...
 		HandlerFunc(utils.WrapHandlerFunc(t.handleSendTransaction))
 	sub.Path("/{id}").
 		Methods(http.MethodGet).
-		Name("transactions_get_tx").
+		Name("GET /transactions/${id}").
 		HandlerFunc(utils.WrapHandlerFunc(t.handleGetTransactionByID))
 	sub.Path("/{id}/receipt").
 		Methods(http.MethodGet).
-		Name("transactions_get_receipt").
+		Name("GET /transactions/${id}/receipt").
```
paologalligit, line 229: Here the variable is in the format `${id}` while [here](https://github.com/vechain/thor/blob/24b7f907b3cbec96c0dbfe75e4104a15bf373a43/api/blocks/blocks.go#L98) it's just `{revision}` without the dollar. I know it's just a string name but I think it's better to have a consistent naming convention

paologalligit (approved):
Looks good, just left a minor comment

api/transactions/transactions.go
darrenvechain: Yepp, you're right, will change this

libotony (approved):

api/metrics.go
```diff
@@ -62,32 +64,32 @@ func metricsMiddleware(next http.Handler) http.Handler {
...
 		}
 
 		next.ServeHTTP(mrw, r)
 
-		if subscription != "" {
-			metricActiveWebsocketCount().AddWithLabel(-1, map[string]string{"subject": subscription})
+		if subscription {
+			metricActiveWebsocketGauge().AddWithLabel(-1, map[string]string{"name": name})
+			// record websocket duration in seconds, not MS
+			metricWebsocketDuration().ObserveWithLabels(time.Since(now).Milliseconds() / 1000, map[string]string{"name": name, "code": strconv.Itoa(mrw.statusCode)})
```
darrenvechain, line 92: This allows us to capture error codes + websocket durations (in seconds)

image

api/metrics.go
```diff
@@ -21,7 +21,9 @@ import (
 var (
 	metricHTTPReqCounter       = metrics.LazyLoadCounterVec("api_request_count", []string{"name", "code", "method"})
 	metricHTTPReqDuration      = metrics.LazyLoadHistogramVec("api_duration_ms", []string{"name", "code", "method"}, metrics.BucketHTTPReqs)
-	metricActiveWebsocketCount = metrics.LazyLoadGaugeVec("api_active_websocket_count", []string{"subject"})
+	metricWebsocketDuration    = metrics.LazyLoadHistogramVec("api_websocket_duration", []string{"name", "code"}, metrics.BucketHTTPReqs)
```
libotony, line 24: The bucket might require a unique one, most wescoket will have a long duration. And is there somewhere to describe this to be in seconds?

darrenvechain: So the current bucket is:

```golang
        BucketHTTPReqs = []int64{
		0, 1, 2, 5, 10, 20, 30, 50, 75, 100,
		150, 200, 300, 400, 500, 750, 1000,
		1500, 2000, 3000, 4000, 5000, 10000,
	}
```

Since we record in seconds, 10000 -> 166 minutes. Anything over this will be recorded in the `Inf` bucket. What timeframe would you suggest to add?

darrenvechain: Could go up to days/ weeks /months I guess

libotony: > Since we record in seconds

darrenvechain: Updated!

metrics/telemetry.go
```diff
@@ -91,13 +98,17 @@ func GaugeVec(name string, labels []string) GaugeVecMeter {
...
+
 // LazyLoad allows to defer the instantiation of the metric while allowing its definition. More clearly:
 // - it allow metrics to be defined and used package wide (using var)
 // - it avoid metrics definition to determine the singleton to use (noop vs prometheus)
 func LazyLoad[T any](f func() T) func() T {
 	var result T
 	var loaded bool
 	return func() T {
+		lazyLoadLock.Lock()
+		defer lazyLoadLock.Unlock()
```
libotony, line 111: Not getting this one. Any examples?

darrenvechain: The websocket duration metric was causing issues when running:

```bash
make test-coverage
```

It seems there was a race condition where the below would be executed twice:

```golang
if !loaded {
	result = f()
	loaded = true
}
```

See this GHA: https://github.com/vechain/thor/actions/runs/12027304858/job/33528077050

darrenvechain: @libotony I was ablo to do:

```go
once.Do(func() {
	result = f()
})
```

libotony (approved):

PR #899, opened 2024-11-29
add 'raw' query parameter to the blocks

YeahNotSewerSide:
# Description

This PR adds a `raw` query parameter to the `/blocks` endpoint, which returns RLP encoded block.

## Type of change

# Checklist:

### DIFF
--- a/api/blocks/blocks.go
+++ b/api/blocks/blocks.go
@@ -37,13 +37,17 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error 
 	if err != nil {
 		return utils.BadRequest(errors.WithMessage(err, "revision"))
 	}
-	raw := req.URL.Query().Get("raw")
-	if raw != "" && raw != "false" && raw != "true" {
-		return utils.BadRequest(errors.WithMessage(errors.New("should be boolean"), "raw"))
+	raw, err := utils.StringToBoolean(req.URL.Query().Get("raw"), false)
+	if err != nil {
+		return utils.BadRequest(errors.WithMessage(err, "raw"))
 	}
-	expanded := req.URL.Query().Get("expanded")
-	if expanded != "" && expanded != "false" && expanded != "true" {
-		return utils.BadRequest(errors.WithMessage(errors.New("should be boolean"), "expanded"))
+	expanded, err := utils.StringToBoolean(req.URL.Query().Get("expanded"), false)
+	if err != nil {
+		return utils.BadRequest(errors.WithMessage(err, "expanded"))
+	}
+
+	if raw && expanded {
+		return utils.BadRequest(errors.WithMessage(errors.New("Raw and Expanded are mutually exclusive"), "raw&expanded"))
 	}
 
 	summary, err := utils.GetSummary(revision, b.repo, b.bft)
@@ -54,7 +58,7 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error 
 		return err
 	}
 
-	if raw == "true" {
+	if raw {
 		rlpEncoded, err := rlp.EncodeToBytes(summary.Header)
 		if err != nil {
 			return err
@@ -78,7 +82,7 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error 
 	}
 
 	jSummary := buildJSONBlockSummary(summary, isTrunk, isFinalized)
-	if expanded == "true" {
+	if expanded {
 		txs, err := b.repo.GetBlockTransactions(summary.Header.ID())
 		if err != nil {
 			return err
--- a/api/blocks/blocks.go
+++ b/api/blocks/blocks.go
@@ -6,8 +6,11 @@
 package blocks
 
 import (
+	"encoding/hex"
+	"fmt"
 	"net/http"
 
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/gorilla/mux"
 	"github.com/pkg/errors"
 	"github.com/vechain/thor/v2/api/utils"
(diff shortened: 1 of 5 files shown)


api/blocks/blocks.go
```diff
@@ -47,6 +54,16 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error
 		return err
 	}
 
+	if raw == "true" {
+		rlpEncodedSummary, err := rlp.EncodeToBytes(summary)
```
libotony, line 58:
```suggestion
		rlpEncoded, err := rlp.EncodeToBytes(summary.Header)
```
Here we should use Header.

api/blocks/blocks.go
```diff
@@ -47,6 +54,16 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error
 		return err
 	}
 
+	if raw == "true" {
```
libotony, line 57: Should also consider that the priority of `raw` is higher than `expanded`.

api/blocks/blocks.go
```diff
@@ -47,6 +54,16 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error
...
+		rlpEncodedSummary, err := rlp.EncodeToBytes(summary)
+		if err != nil {
+			return err
+		}
+		return utils.WriteJSON(w, &JSONRawBlockSummary{
```
darrenvechain, line 62: Can we create a unit test to make the API call and decode the response body?

api/blocks/blocks.go
```diff
@@ -34,6 +37,10 @@ func (b *Blocks) handleGetBlock(w http.ResponseWriter, req *http.Request) error
 	if err != nil {
 		return utils.BadRequest(errors.WithMessage(err, "revision"))
 	}
+	raw := req.URL.Query().Get("raw")
```
darrenvechain, line 40: Should be throw a 400 if raw = true and expanded = true?

api/blocks/blocks.go
YeahNotSewerSide: `raw` is already taking precedence over `expanded`, should it be the other way around?
 https://github.com/vechain/thor/blob/1f79185698a38845980de1a8c2a8b0ff99f143d2/api/blocks/blocks.go#L57

api/blocks/blocks.go
YeahNotSewerSide: As @libotony suggested I made the `raw` to return summary.Header, instead of summary, should I return the whole summary if `expanded = true`?

libotony: Summary doesn't play a role of 'expanded' block, it is designed for internal storage structure.

api/blocks/blocks.go
libotony: My bad, you're right

api/blocks/blocks.go
leszek-vechain: was going to ask about tests - do we need also to update e2e tests for it ?

YeahNotSewerSide: Yeah, I think so and also internal unit tests in thor

YeahNotSewerSide: https://github.com/vechain/thor/pull/899/commits/dc802e2fed08c72cd5f6290ae5ff8b77f7165a43

api/blocks/blocks_test.go
```diff
@@ -67,6 +71,22 @@ func testBadQueryParams(t *testing.T) {
...
+	badQueryParams = "?raw=1"
+	res, statusCode, err = tclient.RawHTTPClient().RawHTTPGet("/blocks/best" + badQueryParams)
+	require.NoError(t, err)
+
+	assert.Equal(t, http.StatusBadRequest, statusCode)
+	assert.Equal(t, "raw: should be boolean", strings.TrimSpace(string(res)))
+}
+
+func testMutuallyExclusiveQueries(t *testing.T) {
+	badQueryParams := "?expanded=true&raw=true"
```
leszek-vechain, line 84: maybe we can add also combinations here:

- expanded = true, raw. = false
- expanded = false, raw = true
- expanded = false, raw = false

YeahNotSewerSide: @leszek-vechain I have checked the code, there are already checks for that, was that what you mean?
https://github.com/vechain/thor/blob/96634c89c54c4712384c6d00ecf955e03e2ca345/api/blocks/blocks_test.go#L84
https://github.com/vechain/thor/blob/96634c89c54c4712384c6d00ecf955e03e2ca345/api/blocks/blocks_test.go#L93
https://github.com/vechain/thor/blob/96634c89c54c4712384c6d00ecf955e03e2ca345/api/blocks/blocks_test.go#L104
https://github.com/vechain/thor/blob/96634c89c54c4712384c6d00ecf955e03e2ca345/api/blocks/blocks_test.go#L194

api/utils/http.go
```diff
@@ -36,6 +38,17 @@ func BadRequest(cause error) error {
 	}
 }
 
+func StringToBoolean(boolStr string, defaultVal bool) (bool, error) {
+	if boolStr == "" {
+		return defaultVal, nil
+	} else if boolStr == "false" {
+		return false, nil
+	} else if boolStr == "true" {
```
otherview, line 46: Given it's a direct comparison, would it be useful to use string trim ?

otherview (approved):

api/blocks/blocks_test.go
leszek-vechain: yeah nice :)

api/utils/http.go
YeahNotSewerSide: Yeah, it might be helpful, will add that

YeahNotSewerSide: I just checked I don't think other endpoints use trimming for parameters, not sure now if there has to be this inconsistency

libotony (approved):

PR #900, opened 2024-11-29
bug: fix logs endpoints query

darrenvechain:
# Description

This PR changes the logs `Range` to use pointers so that `to` is not set as 0 when omitted. Reproduce the issue here:

```bash
curl --request POST \
  --url https://mainnet.vechain.org/logs/event \
  --header 'Accept: application/json, text/plain' \
  --header 'Content-Type: application/json' \
  --data '{
  "range": {
    "unit": "block",
    "from": 0
  },
  "options": {
    "offset": 0,
    "limit": 100
  },
  "criteriaSet": [
    {
      "address": "0x0000000000000000000000000000456E65726779",
      "topic0": "0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef",
      "topic1": "0x0000000000000000000000006d95e6dca01d109882fe1726a2fb9865fa41e7aa"
    }
  ],
  "order": "asc"
}'
```

# Checklist:


api/events/types.go
```diff
@@ -129,14 +129,30 @@ const (
 
 type Range struct {
 	Unit RangeType
-	From uint64
-	To   uint64
+	From *uint64 `json:"from,omitempty"`
+	To   *uint64 `json:"to,omitempty"`
+}
+
+func NewRange(unit RangeType, from uint64, to uint64) *Range {
```
libotony, line 136: Seems test only?

darrenvechain: Yepp, its just annoying to do this everytime:

```golang
to := 0
from := 1

range := &Range{
  from: &from
  to: &to
}
```

libotony: How about put the func to test file?

api/events/types.go
```diff
@@ -129,14 +129,22 @@ const (
...
 	if r == nil {
 		return nil, nil
 	}
+	if r.From == nil {
+		from := uint64(0)
+		r.From = &from
+	}
+	if r.To == nil {
+		to := uint64(math.MaxUint64)
+		r.To = &to
```
otherview: Is it correct to say that previous if the to was nil it would be set to 0 and now it will be set to MaxUint64 ? 
If so, could that create some issue ?

darrenvechain: It depends.

If the `range` object was completely omitted, the entire chain would be queried.

If the `range` was set, but not the `to`, it would default to 0, which is different behaviour to the above. So this sets the `to` to the max possible. But it gets set to the max of that and the chain head below

otherview: Yeah, that's a small wrinkle, but I think that's fine. No massive change in expected behaviour.

libotony (approved):

PR #907, opened 2024-12-05
chore: make thorclient configurable

darrenvechain:
# Description

Make thorclient more configurable my specifying `*http.Client` 

# Checklist:

### DIFF
--- a/api/accounts/accounts_test.go
+++ b/api/accounts/accounts_test.go
@@ -106,7 +106,7 @@ func TestAccount(t *testing.T) {
 	initAccountServer(t, true)
 	defer ts.Close()
 
-	tclient = thorclient.New(ts.URL)
+	tclient = thorclient.New(ts.URL, &http.Client{})
 	for name, tt := range map[string]func(*testing.T){
 		"getAccount":                          getAccount,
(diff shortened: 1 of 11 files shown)


otherview (commented):
I think this is the right path forward, the http object is always instantiated without params in the `New()`.

To avoid having `client := New(ts.URL, &http.Client{})` everywhere, would it make sense to have `client := New(ts.URL)` (which has the default httpclient) and `client := NewWithHTTP(ts.URL, &http.Client{})` ?

darrenvechain:
> I think this is the right path forward, the http object is always instantiated without params in the `New()`.
> 
> To avoid having `client := New(ts.URL, &http.Client{})` everywhere, would it make sense to have `client := New(ts.URL)` (which has the default httpclient) and `client := NewWithHTTP(ts.URL, &http.Client{})` ?

Ya I had considered this, but my reasoning was that as a client of the codebase you probably only have to init the client once and pass the struct around your app.

We only have a lot of `&http.Client{}` because we use it a lot in our tests, and IMO its not that hard to initialise it.

I don't feel strongly either way

PR #916, opened 2024-12-10
Metrics (Accounts): State changes

freemanzMrojo:
# Description

This PR adds metrics for accounts in the state.


state/state.go
```diff
@@ -538,6 +540,8 @@ func (s *State) Stage(newVer trie.Version) (*Stage, error) {
 					return err
 				}
 			}
+			// Just once for the account trie.
```
libotony, line 543: I would suggest avoid including metrics in state, as it's the primitive operation and all the metrics will be fallen into trie node cache.

chain/cache.go
```diff
@@ -20,6 +20,7 @@ func newCache(maxSize int) *cache {
 
 func (c *cache) GetOrLoad(key interface{}, load func() (interface{}, error)) (interface{}, error) {
 	if value, ok := c.Get(key); ok {
+		metricBlockRepositoryCounter().AddWithLabel(1, map[string]string{"type": "read", "target": "cache"})
```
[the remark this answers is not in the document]
freemanzMrojo: These metrics are per object, as described in the title. I do not know why #910 was done since cache miss/rate was already covered [by this](https://github.com/vechain/thor/blob/9d359f702b32712cf8acd4a4fc53393419509c35/muxdb/internal/trie/metrics.go#L9).

But it is true that [this other cache](https://github.com/vechain/thor/blob/9d359f702b32712cf8acd4a4fc53393419509c35/chain/metric.go#L9) looks like a duplicate since it is placed precisely in receipts, transactions and so on with the difference that only applies to reads.

I can align both things. Wdyt?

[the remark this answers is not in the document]
darrenvechain: I would have put this metric here, but I also wanted to record the domain cache/hit, ie `blocks`, `transactions` or `receipts`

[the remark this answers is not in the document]
libotony: After looking into the names, now I understand the purpose of them. Regarding `Reads`, I would suggest leverage `cache.hit` + `cache.miss`.  Regarding `Writes` I would consider they are immutable based on the blocks imported. Is write inevitable from your point of view?

state/state.go
libotony: Thank you for the explanation, I was proposed avoiding including metrics in the state package, reason for that is avoiding potential huge overhead and all state operations are trie operations. Since we already got metrics in the trie package, why don't leave state package without metrics. 

Secondly, commit function usually being called for every block processing, I would suggest remove this metric.

state/state.go
```diff
@@ -131,6 +132,7 @@ func (s *State) getAccount(addr thor.Address) (*Account, error) {
 	if err != nil {
 		return nil, err
 	}
+	metricAccountCounter().AddWithLabel(1, map[string]string{"type": "read", "target": "stackedmap"})
```
libotony, line 135: The `stackedmap` is an in-memory cache stuff with stack functionality, I don't see a necessity of adding metrics for it.

state/state.go
```diff
@@ -120,6 +120,7 @@ func (s *State) getCachedObject(addr thor.Address) (*cachedObject, error) {
 	if err != nil {
 		return nil, err
 	}
+	metricAccountCounter().AddWithLabel(1, map[string]string{"type": "read", "target": "trie"})
```
libotony, line 123: `State.GetAccount` -> `stackedMap.Get` -> `state.getCachedObject` -> `state.LoadAccount` -> `trie.Get`

This might not be reflecting the real calling path, but just for demonstrating the path from state to the trie package. I would suggest consider in the low-level packages other than this one.

freemanzMrojo: The problem of this approach is that we miss the concept of object, so at  the `trie` level we do not know whether is an `account` or something else. That method is actually used also by the `chain` package.

state/state.go
freemanzMrojo: Ok, will get rid of the `stackedmap` metrics then, thanks.

state/state.go
freemanzMrojo: The idea of having these metrics here is because of the object, in this case `accounts`. At `trie` we do not have that visibility, hence we cannot figure this out.

chain/cache.go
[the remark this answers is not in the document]
freemanzMrojo: It is indeed a duplicate, the thing is I was working on this ticket so as you can imagine I was not expecting someone else doing the same thing.

About `Writes`, you mean block-wise or also for the other "objects" (tx, receipts...)? Regarding blocks, the metric is only submitted to Prometheus when calling `AddBlock`. Not sure what you mean by "inevitable", should not we record them?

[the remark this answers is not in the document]
libotony: Nah, I might use wrong word regarding 'inevitable', I was thinking that writing to block/transaction/receipt was deterministic.

state/state.go
libotony: Let's solve this offline.

metrics/prometheus.go
```diff
@@ -204,7 +204,7 @@ func (o *prometheusMetrics) newHistogramMeter(name string, buckets []int64) Hist
 
 	err := prometheus.Register(meter)
 	if err != nil {
-		logger.Warn("unable to register metric", "err", err)
+		logger.Warn(fmt.Sprintf("unable to register metric %s", name), "err", err)
```
otherview, line 207: Small nit - usage of named vars in the logger:
```suggestion
		logger.Warn("unable to register metric", "metric", name, "err", err)
```

freemanzMrojo:
Scope changed after conversation with @otherview, changes made accordingly.

otherview:
My bad, this should have made it to the RC. Will add it to the RC-2 👍

darrenvechain (approved):

PR #926, opened 2024-12-18
Refactor the subscriptions api to meet standard mount approach

otherview:
# Description

This ticket refactors the `handlesubject` into the `websocked` method.

Fixes # [(issue)](https://github.com/vechain/otherviews-workboard/issues/115)

# Checklist:

### DIFF
--- a/api/subscriptions/subscriptions.go
+++ b/api/subscriptions/subscriptions.go
@@ -386,32 +386,32 @@ func (s *Subscriptions) Mount(root *mux.Router, pathPrefix string) {
 
 	sub.Path("/txpool").
 		Methods(http.MethodGet).
-		Name("subscriptions_pending_tx").
+		Name("WS subscriptions/txpool").
 		HandlerFunc(utils.WrapHandlerFunc(s.handlePendingTransactions))
 
 	sub.Path("/block").
 		Methods(http.MethodGet).
-		Name("subscriptions_block").
+		Name("WS subscriptions/block").
 		HandlerFunc(utils.WrapHandlerFunc(s.websocket(s.handleBlockReader)))
 
 	sub.Path("/event").
 		Methods(http.MethodGet).
-		Name("subscriptions_event").
+		Name("WS subscriptions/event").
 		HandlerFunc(utils.WrapHandlerFunc(s.websocket(s.handleEventReader)))
 
 	sub.Path("/transfer").
 		Methods(http.MethodGet).
-		Name("subscriptions_transfer").
+		Name("WS subscriptions/transfer").
 		HandlerFunc(utils.WrapHandlerFunc(s.websocket(s.handleTransferReader)))
 
 	sub.Path("/beat2").
 		Methods(http.MethodGet).
-		Name("subscriptions_beat2").
+		Name("WS subscriptions/beat2").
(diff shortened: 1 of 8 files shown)


api/subscriptions/subscriptions_test.go
```diff
@@ -59,7 +59,7 @@ func TestDeprecatedSubscriptions(t *testing.T) {
 
 	_, resp, err := websocket.DefaultDialer.Dial(u.String(), nil)
 	assert.Error(t, err)
-	assert.Equal(t, http.StatusGone, resp.StatusCode)
+	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
```
paologalligit, line 62: Can we leave the gone status like before? It looks more precise when telling something that was there before but now it's been deprecated

otherview: Yep, good call, fixed 👍

api/subscriptions/subscriptions.go
```diff
@@ -387,15 +339,83 @@ func (s *Subscriptions) Close() {
...
-	sub.Path("/{subject:beat|beat2|block|event|transfer}").
+
+	sub.Path("/block").
+		Methods(http.MethodGet).
+		Name("subscriptions_block").
+		HandlerFunc(utils.WrapHandlerFunc(s.websocket(s.handleBlockReader)))
+
+	sub.Path("/event").
+		Methods(http.MethodGet).
+		Name("subscriptions_event").
```
darrenvechain, line 399: @otherview could we update these to same pattern as other APIs now? I think we didn't do it previously because it was one mount for all endpoints.

Eg.

```
Name("WS /subsciptions/beat")
```

One thing to look out for would be the metrics. I think for subs we expect the name to start with subscription

otherview: Good idea! Fixed! Looks good now. 
image

darrenvechain (dismissed):

api/subscriptions/subscriptions.go
```diff
@@ -387,15 +339,83 @@ func (s *Subscriptions) Close() {
...
+		err = s.pipe(conn, reader, closed)
+		if err != nil {
+			logger.Debug("error in websocket pipe", "err", err)
+		}
+		return err
+	}
+}
+
+// handleGone is a handler for deprecated endpoints that returns HTTP 410 Gone.
+func handleGone(w http.ResponseWriter, _ *http.Request) error {
```
libotony, line 378: Can be used in other deprecated endpoints? If so, maybe move it to utils folder?

libotony:
Appended a minor naming commit @otherview

api/subscriptions/subscriptions.go
```diff
@@ -387,15 +339,76 @@ func (s *Subscriptions) Close() {
...
+
+		// Call the provided reader function
+		reader, err := readerFunc(w, req)
+		if err != nil {
+			return err
+		}
+
+		// Ensure cleanup when the connection closes
+		defer func() {
+			if closer, ok := reader.(interface{ Close() }); ok {
```
libotony, line 355: `MsgReader` doesn't required to implement a `closer` interface, this code block is not required.

otherview: Oh yeah. Leftover from the handlePendingTransactions refactor. Good catch! 🙏

libotony (approved):

PR #928, opened 2024-12-19
Use thread safe LRU implementation

YeahNotSewerSide:
# Description

Replace thread unsafe LRU cache with a thread safe version.
I have decided to use the third-party library instead of priority cache, as this implementation suites better for these needs and it's optimized with RWMutex.

Fixes # [(issue)](https://github.com/vechain/protocol-board-repo/issues/373)

## Type of change

# How Has This Been Tested?

Ran tests on thor.
Ran a PoC made to test this issue.
Tested beat2 endpoint.

# Checklist:

### DIFF
--- a/api/subscriptions/message_cache.go
+++ b/api/subscriptions/message_cache.go
@@ -7,7 +7,8 @@ package subscriptions
 
 import (
 	"fmt"
-	"github.com/hashicorp/golang-lru/v2"
+
+	lru "github.com/hashicorp/golang-lru/v2"
 	"github.com/vechain/thor/v2/thor"
 )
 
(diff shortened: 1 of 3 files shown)


darrenvechain (dismissed):

api/subscriptions/message_cache.go
```diff
@@ -41,16 +41,16 @@ func newMessageCache[T any](cacheSize uint32) *messageCache[T] {
...
 	if ok {
 		return msg.(T), false, nil
 	}
 
-	mc.mu.Lock()
-	defer mc.mu.Unlock()
-	msg, ok = mc.cache.Get(id)
+	mc.w.Lock()
+	defer mc.w.Unlock()
+	msg, ok = mc.cache.Peek(id)
```
leszek-vechain, line 53: is it possible to add some tests around this new method ? or the old ones will still be enough. As I understand we cant add the actual reproduce method - it is a bit tricky

YeahNotSewerSide: I don't think it's possible to create a test that will be giving errors 100% of the time, when there is an actual problem.

libotony (approved):

darrenvechain (approved):

PR #934, opened 2025-01-10
Metrics (Block processing lock): Number of times + duration

freemanzMrojo:
# Description

Metrics taken based on the description for `Block processing lock` here https://github.com/vechain/protocol-board-repo/issues/340 and taking `metricBlockProcessedCount` and `metricBlockProcessedDuration` as a reference.

### DIFF
--- a/cmd/thor/node/metrics.go
+++ b/cmd/thor/node/metrics.go
@@ -10,10 +10,12 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


libotony:
As block processing duration is already captured, I would suggest this metric to be redundant.

otherview:
> As block processing duration is already captured, I would suggest this metric to be redundant.

Agree, perhaps this metrics doesn't give us an extra data point that would justify it ?

PR #936, opened 2025-01-10
Metrics (Compaction): Values by level

freemanzMrojo:
# Description

The frequency of the compaction as such is not exposed by the leveldb library but it returns a string with these compactions stats (following this https://github.com/vechain/protocol-board-repo/issues/340 ).

The code in this PR formats the table and extracts the values (tables, sizes, writes, reads, time taken by the compaction) by level.

This metric is collected every 10 seconds.


muxdb/metrics.go
```diff
@@ -8,7 +8,136 @@
...
+	// Round the float64 value
+	roundedValue := math.Round(floatValue)
+
+	// Convert the rounded float64 to int64
+	intValue := int64(roundedValue)
+
+	return intValue, nil
+}
+
+func extractCompactionValues(stats string) ([]CompactionValues, error) {
```
otherview, line 92: What if an upgrade was to happen and the output string would change? 
It does look like an error would be caught in the `parseCompactionColumns` but I wonder if it might be missed until prod ? Perhaps a unit test to keep it tight could be helpful.

muxdb/metrics.go
```diff
@@ -8,7 +8,136 @@
...
+	// Print the stats string using the tabwriter
+	fmt.Fprintln(w, stats)
+	w.Flush()
+
+	// Extract and log the value from the specified level
+	formattedStats := sb.String()
+	logger.Debug(formattedStats)
+	values, err := extractCompactionValues(formattedStats)
+	if err != nil {
+		logger.Error("Failed to extract values for stats %s: %v", stats, err)
```
otherview, line 64:
```suggestion
		logger.Error("Failed to extract values for stats %s: %s", stats, err.Error())
```

muxdb/metrics.go
```diff
@@ -8,7 +8,136 @@
...
+	}
+
+	if len(values) == 0 {
+		return nil, fmt.Errorf("no valid compaction values found in stats %s", stats)
+	}
+
+	return values, nil
+}
+
+func parseCompactionColumns(columns []string) (CompactionValues, error) {
```
otherview, line 114: A pointer ref should be ok.

```suggestion
func parseCompactionColumns(columns []string) (*CompactionValues, error) {
```

muxdb/metrics.go
```diff
@@ -8,7 +8,136 @@
...
+	tables, err := strconv.ParseInt(columns[2], 10, 64)
+	if err != nil {
+		return CompactionValues{}, fmt.Errorf("error when parsing tables: %v", err)
```
otherview, line 117:
```suggestion
		return nil, fmt.Errorf("error when parsing tables: %v", err)
```

muxdb/muxdb.go
```diff
@@ -62,6 +63,21 @@ type MuxDB struct {
 	trieBackend *backend
 }
 
+func collectCompactionMetrics(ldb *leveldb.DB) {
+	ticker := time.NewTicker(10 * time.Second)
+	defer ticker.Stop()
+
+	for range ticker.C {
```
otherview, line 70: Should this process also stop when the muxdb is `Close()` ?

freemanzMrojo: Modification done here https://github.com/vechain/thor/pull/936/commits/916c017ca6447069dedd15fb96f7383b2bce1f1c to allow that, please have a look, thanks.

muxdb/metrics.go
freemanzMrojo: Tests added in the last commit, thanks.

otherview (approved):

PR #937, opened 2025-01-11
Metrics (Batch writes performance): Bytes written per bulk + duration

freemanzMrojo:
# Description

This PR covers `Batch Write Performance` from this ticket https://github.com/vechain/protocol-board-repo/issues/340.

Since the leveldb library as such does not return info about this, we set the metrics at the operation level within `engine`.

### DIFF
--- a/metrics/telemetry.go
+++ b/metrics/telemetry.go
@@ -30,6 +30,11 @@ func HTTPHandler() http.Handler {
 	return metrics.GetOrCreateHandler()
 }
 
+func NoOp() bool {
+	_, isNoOp := metrics.(*noopMetrics)
+	return isNoOp
+}
+
 // Define standard buckets for histograms
 var (
 	Bucket10s      = []int64{0, 500, 1000, 2000, 3000, 4000, 5000, 7500, 10_000}
--- a/muxdb/muxdb.go
+++ b/muxdb/muxdb.go
@@ -20,6 +20,7 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/storage"
 	"github.com/vechain/thor/v2/kv"
 	"github.com/vechain/thor/v2/log"
+	"github.com/vechain/thor/v2/metrics"
 	"github.com/vechain/thor/v2/muxdb/engine"
 	"github.com/vechain/thor/v2/trie"
 )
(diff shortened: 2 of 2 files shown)


muxdb/engine/leveldb.go
```diff
@@ -112,14 +114,20 @@ func (ldb *levelEngine) Bulk() kv.Bulk {
...
-					return err
+		if batch != nil {
+			if batchBytesNo := len(batch.Dump()); batchBytesNo >= minSize {
+				if batch.Len() > 0 {
+					startTime := mclock.Now()
+					if err := ldb.db.Write(batch, &writeOpt); err != nil {
+						return err
+					}
+					metricBatchWriteBytes().Set(int64(batchBytesNo))
+					batchWriteElapsed := mclock.Now() - startTime
```
YeahNotSewerSide, line 125: Hey, does it make sense to change these lines order?
```suggestion
batchWriteElapsed := mclock.Now() - startTime
metricBatchWriteBytes().Set(int64(batchBytesNo))
```

freemanzMrojo: They both get different values (`batchBytesNo` for the number of bytes and `batchWriteElapsed` for the time) so IMO it does not make a difference since the event has finished in any case.

But anyways, I am going to accept your suggestion. Thanks for reviewing!

muxdb/engine/leveldb.go
```diff
@@ -112,14 +114,20 @@ func (ldb *levelEngine) Bulk() kv.Bulk {
...
+		if batch != nil {
+			if batchBytesNo := len(batch.Dump()); batchBytesNo >= minSize {
+				if batch.Len() > 0 {
+					startTime := mclock.Now()
+					if err := ldb.db.Write(batch, &writeOpt); err != nil {
+						return err
+					}
+					batchWriteElapsed := mclock.Now() - startTime
+					metricBatchWriteBytes().Set(int64(batchBytesNo))
+					metricBatchWriteDuration().Observe(time.Duration(batchWriteElapsed).Milliseconds())
```
otherview, line 126: Is the mclock is required ?
I think the `time` package, `time.Now()` and `time.Since()` should work well.

freemanzMrojo: replace by `time`, thanks.

libotony:
FYI as we have discussed offline, https://github.com/vechain/goleveldb/blob/master/leveldb/db.go#L1026-L1050

PR #952, opened 2025-01-20
Metrics (Compaction): Do not call the DB if metrics are disabled

freemanzMrojo:
# Description

This PR makes sure that we just call the db to get stats if the metrics are enabled (so different to noop).

### DIFF
--- a/muxdb/muxdb.go
+++ b/muxdb/muxdb.go
@@ -73,7 +73,7 @@ func collectCompactionMetrics(ctx context.Context, ldb *leveldb.DB) {
 		// We avoid calling the db if metrics are disabled
 		return
 	}
-	
+
 	ticker := time.NewTicker(10 * time.Second)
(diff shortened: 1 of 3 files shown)


darrenvechain (approved):

darrenvechain (approved):

darrenvechain (approved):

muxdb/muxdb.go
```diff
@@ -68,6 +69,11 @@ type MuxDB struct {
 
 // collectCompactionMetrics collects compaction metrics periodically.
 func collectCompactionMetrics(ctx context.Context, ldb *leveldb.DB) {
+	if metrics.NoOp() {
```
otherview, line 72: We want to avoid these types of conditional behaviours. 
If the metrics are not instantiated, it should use the noopMetrics (`noop.go`) implementation and not produce metrics.

PR #956, opened 2025-01-23
feat(solo): keep thor solo genesis ID consistent for all forks

darrenvechain:
# Description

Define a new `ForkConfig` for solo so the genesis ID does not change.

# Checklist:

### DIFF
--- a/api/subscriptions/pending_tx_test.go
+++ b/api/subscriptions/pending_tx_test.go
@@ -71,7 +71,7 @@ func TestPendingTx_Unsubscribe(t *testing.T) {
 
 func TestPendingTx_DispatchLoop(t *testing.T) {
 	db := muxdb.NewMem()
-	gene := genesis.NewDevnet()
+	gene := genesis.NewDevnet(thor.SoloFork)
 	stater := state.NewStater(db)
 	b0, _, _, _ := gene.Build(stater)
 	repo, _ := chain.NewRepository(db, b0)
--- a/api/subscriptions/pending_tx_test.go
+++ b/api/subscriptions/pending_tx_test.go
@@ -71,7 +71,7 @@ func TestPendingTx_Unsubscribe(t *testing.T) {
 
 func TestPendingTx_DispatchLoop(t *testing.T) {
 	db := muxdb.NewMem()
-	gene := genesis.NewDevnet(thor.SoloFork)
+	gene := genesis.NewDevnet()
 	stater := state.NewStater(db)
 	b0, _, _, _ := gene.Build(stater)
 	repo, _ := chain.NewRepository(db, b0)
--- a/cmd/thor/main.go
+++ b/cmd/thor/main.go
@@ -170,8 +170,9 @@ func defaultAction(ctx *cli.Context) error {
 	logLevel := initLogger(lvl, ctx.Bool(jsonLogsFlag.Name))
 
 	// enable metrics as soon as possible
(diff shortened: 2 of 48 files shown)


api/subscriptions/pending_tx_test.go
```diff
@@ -71,7 +71,7 @@ func TestPendingTx_Unsubscribe(t *testing.T) {
 
 func TestPendingTx_DispatchLoop(t *testing.T) {
 	db := muxdb.NewMem()
-	gene := genesis.NewDevnet()
+	gene := genesis.NewDevnet(thor.SoloFork)
```
leszek-vechain, line 74: why this one is SoloFork ?

darrenvechain: `genesis.NewDevnet(thor.SoloFork)`  now equals `genesis.NewDevnet()` previously

cmd/thor/main.go
```diff
@@ -338,13 +338,12 @@ func soloAction(ctx *cli.Context) error {
 
 	var (
 		gene       *genesis.Genesis
-		forkConfig thor.ForkConfig
+		forkConfig = thor.SoloFork
 	)
 
 	flagGenesis := ctx.String(genesisFlag.Name)
 	if flagGenesis == "" {
-		gene = genesis.NewDevnet()
```
libotony, line 346: How about let `genesis.NewDevnet()` returning the `devForkConfig`?

libotony:
I think this solves most scenarios, except the existing solo instance with `--persist` flag set. This might be a complicated use case, I'll let you make a decision, document that scenario or not(considering it to be complicated).

cmd/thor/main.go
darrenvechain: I think its very useful that `NewDevnet` accepts the fork config. It allow us to integration test forks

darrenvechain:
> I think this solves most scenarios, except the existing solo instance with `--persist` flag set. This might be a complicated use case, I'll let you make a decision, document that scenario or not(considering it might bring more questions).

I don't think we can actually solve this one without doing some hacky checks in `solo.go`

cmd/thor/solo/solo.go
```diff
@@ -237,8 +237,8 @@ func (s *Solo) init(ctx context.Context) error {
 		return err
 	}
 
-	if !s.onDemand {
-		// wait for the next block interval if not on-demand
+	// if we're at the genesis block, instantly perform all forks at block 1, else wait for next block time
```
libotony, line 240: This might break non on-demand mode?

darrenvechain: The idea is that block 1 is always instantly minted.

This if condition more so when a previous state exists with `--persist`, the we wait for the regular block flow to update the base gas price

test/testchain/chain.go
```diff
@@ -64,15 +64,15 @@ func New(
 
 // NewIntegrationTestChain is a convenience function that creates a Chain for testing.
 // It uses an in-memory database, development network genesis, and a solo BFT engine.
-func NewIntegrationTestChain() (*Chain, error) {
+func NewIntegrationTestChain(config thor.ForkConfig) (*Chain, error) {
```
otherview, line 67: All of NewIntegrationTestChain are using the thor.ForkConfig. Likely ok to hardcode that in?

darrenvechain: Removed it!

PR #965, opened 2025-01-28
feat: `fees/history` `GET` endpoint

freemanzMrojo:
# Description

Implements a version of the `eth_feeHistory` endpoint ([see this](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_feehistory/) as a reference) with the following changes:

- Path: The path is `fees/history` and it is a `GET` endpoint.

- Request (query params):
  1. `newestBlock` is actually our `revision`
  2. There is no `rewardPercentiles`
  3. `blockCount` is a number rather than an hex-encoded uint
  
- Response: 
  1. We return just `baseFees` and `gasUsedRatios` (we keep the types for both of them, so hex and float respectively)
  2. `oldestBlock` is a VeChain block ID instead of an hex-encoded uint 
 
 
 Closes https://github.com/vechain/protocol-board-repo/issues/389

## Release Notes

- **New Features**
    - Added a new `/fees/history` API endpoint to retrieve historical fee data.
    - Introduced a fee caching mechanism for efficient fee information retrieval.
    - Enhanced fee management functionalities within the blockchain application.

- **Improvements**
    - Enhanced API documentation with new fee-related endpoints.
    - Updated `--api-backtrace-limit` flag to support fees APIs.
    - Improved error handling for fee history requests.
    - Added convenience methods for HTTP error responses.
    - Enhanced testing framework for fee-related functionalities.
    - Modified integration test chain creation for flexible configuration.

- **Documentation**
    - Updated usage documentation to reflect new fee API capabilities.


api/fees/fees.go
```diff
@@ -0,0 +1,158 @@
...
+func (f *Fees) validateGetFeesHistoryParams(req *http.Request) (uint32, *chain.BlockSummary, error) {
+	blockCountParam := req.URL.Query().Get("blockCount")
+	blockCount, err := strconv.ParseUint(blockCountParam, 10, 32)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "invalid blockCount, it should represent an integer"))
+	}
+	if blockCount < 1 || blockCount > maxNumberOfBlocks {
+		return 0, nil, utils.BadRequest(errors.New(fmt.Sprintf("blockCount must be between 1 and %d", maxNumberOfBlocks)))
+	}
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
```
darrenvechain, line 45: Does it make sense to use `revision` as the parameter here to be consistent with other APIs? 

If this is the name in geth, then maybe not

api/fees/fees.go
```diff
@@ -0,0 +1,158 @@
...
+
+	oldestBlock, blockDataChan := f.processBlockRange(blockCount, summary)
+
+	var (
+		baseFeesWithNil = make([]*hexutil.Big, blockCount)
+		gasUsedRatios   = make([]float64, blockCount)
+	)
+
+	// Collect results from the channel
+	for blockData := range blockDataChan {
```
darrenvechain, line 122: Should we use some kind of cache (mabye [PrioCache](https://github.com/vechain/thor/blob/c5d4a14418c6ce72ffacea0a4dd6926a5e8ce142/cache/prio_cache.go#L15)) to store fee history for `maxNumberOfBlocks` ?

api/fees/fees.go
freemanzMrojo: It is the name in the Ethereum implementations, yep. `revision` is confusing too, I had an old implementation with that and I found it confusing tbh (+ the other argument).

api/fees/fees.go
freemanzMrojo: I personally do not think so since the main retriever, `GetSummary`, works also on a cache. 

On top of that, the number of blocks is limited to 1024 which is pretty small IMO.

But it could be done if you guys agree on that.

api/fees/fees_test.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/vechain/thor/v2/thorclient"
+	"github.com/vechain/thor/v2/tx"
+)
+
+func TestFees(t *testing.T) {
+	ts := initFeesServer(t)
+	defer ts.Close()
+
+	tclient := thorclient.New(ts.URL)
+	for name, tt := range map[string]func(*testing.T, *thorclient.Client){
```
paologalligit, line 31: Can you add a test having a `blockCount` larger than the max allowed?

api/fees/fees.go
```diff
@@ -0,0 +1,158 @@
...
+		if blockData.err != nil {
+			return blockData.err
+		}
+		// Ensure the order of the baseFees and gasUsedRatios is correct
+		blockPosition := blockData.blockSummary.Header.Number() - oldestBlock
+		if baseFee := blockData.blockSummary.Header.BaseFee(); baseFee != nil {
+			baseFeesWithNil[blockPosition] = (*hexutil.Big)(baseFee)
+		} else {
+			baseFeesWithNil[blockPosition] = (*hexutil.Big)(big.NewInt(0))
+			fmt.Printf("LLEGA blockPosition: %+v\n", baseFeesWithNil[blockPosition])
```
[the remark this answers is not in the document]
freemanzMrojo: Si, gracias!

api/fees/fees_test.go
freemanzMrojo: Sure thing

api/fees/fees.go
darrenvechain: We might need to benchmark this to know what the best path forward is.

Currently I think the cache size is only 512 (half of limit to query)

api/fees/fees.go
```diff
@@ -0,0 +1,157 @@
...
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "invalid blockCount, it should represent an integer"))
+	}
+	if blockCount < 1 || blockCount > maxNumberOfBlocks {
+		return 0, nil, utils.BadRequest(errors.New(fmt.Sprintf("blockCount must be between 1 and %d", maxNumberOfBlocks)))
+	}
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+	summary, err := utils.GetSummary(newestBlock, f.repo, f.bft)
```
darrenvechain, line 49: Do we need a check for `summary - blockCountParam > 0` ? I'm just thinking in the case of solo

api/fees/fees_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+	router := mux.NewRouter()
+	fees.New(thorChain.Repo(), thorChain.Engine()).
+		Mount(router, "/fees")
+
+	return httptest.NewServer(router)
+}
+
+func getFeeHistory(t *testing.T, tclient *thorclient.Client) {
+	res, statusCode, err := tclient.RawHTTPClient().RawHTTPGet("/fees/history?blockCount=3&newestBlock=best")
```
leszek-vechain, line 83: can we add other revisions too ?

api/fees/fees.go
freemanzMrojo: There is a test (`getFeeHistoryBlockCountZero`) where that case is being validated. Are there any specifics in solo on top of what is tested there?

api/fees/fees.go
freemanzMrojo: IMO would be easier to just increase 512 to 1024 rather than creating another cache that would be duplicating some of the data, like base fees, that are part of the block.

Anyways, since we disagree I think it would be good to have more opinions, I am in for democracy in case a majority prefers another cache 👍

api/fees/fees_test.go
freemanzMrojo: added a test for `finalized`

api/fees/fees.go
darrenvechain: The cache won't be used if the `newestBlock` is lower than the cache (weird use case I know).

I think its still worth benchmarking worst case scenarios. The benchmarks would make the decision for us I guess

freemanzMrojo: What would be repo for those benchmarks? `draupnir` maybe?

darrenvechain: No I mean like a go benchmark:

```
func BenchmarkFees(b *testing.B)
```

Something like create a load of blocks using the `testchain` package and then run queries on the endpoint

freemanzMrojo: The current tests are using `testchain` with 10 blocks (I could create X). Do you think that it would be enough by increasing the limit of that loop so I create 2048 instead of 10 @darrenvechain ?

darrenvechain (dismissed):

api/fees/fees_test.go
```diff
@@ -0,0 +1,198 @@
...
+	tclient = thorclient.New(ts.URL)
+	for name, tt := range map[string]func(*testing.T, *thorclient.Client){
+		"getFeeHistoryWithSummaries": getFeeHistoryWithSummaries,
+		"getFeeHistoryOnlySummaries": getFeeHistoryOnlySummaries,
+	} {
+		t.Run(name, func(t *testing.T) {
+			tt(t, tclient)
+		})
+	}
+}
```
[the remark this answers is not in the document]
freemanzMrojo: Done

api/utils/revisions.go
```diff
@@ -70,17 +70,20 @@ func ParseRevision(revision string, allowNext bool) (*Revision, error) {
...
 }
 
-// GetSummary returns the block summary for the given revision,
-// revision required to be a deterministic block other than "next".
-func GetSummary(rev *Revision, repo *chain.Repository, bft bft.Committer) (sum *chain.BlockSummary, err error) {
+func ParseNumberRevision(revision uint32) *Revision {
+	return &Revision{revision}
+}
+
+func ParseBlockID(rev *Revision, repo *chain.Repository, bft bft.Committer) (thor.Bytes32, error) {
```
otherview, line 77: I was checking the usages of this, and perhaps it might be possible to get away without this new method.

api/api.go
```diff
@@ -97,6 +98,9 @@ func New(
 	subs := subscriptions.New(repo, origins, config.BacktraceLimit, txPool, config.EnableDeprecated)
 	subs.Mount(router, "/subscriptions")
 
+	fees := fees.New(repo, bft, config.BacktraceLimit, 1024)
+	fees.Mount(router, "/fees")
```
otherview, line 102: nit: 
```suggestion
	fees.New(repo, bft, config.BacktraceLimit, 1024).
	Mount(router, "/fees")
```

otherview (commented):
Looks pretty good!

I think it might be possible to simplify the overall logic. Perhaps something that makes responsibilities more decoupled, like :
- Validate and sanitize inputs
- Request data from cache ( assumes requests have previously been validated )
- If the cache doesn't have data, it backfills from the repo

Makefile
```diff
@@ -44,16 +46,46 @@ go_version_check:
...
+	@echo "cleaning build cache and binaries..."
+	@go clean -cache -modcache -i -r
+	@rm -rf $(CURDIR)/bin/*
+	@echo "done. build cache and binaries removed."
+
+test-sequential:| go_version_check #@ Run the tests sequentially for SEQUENTIAL_TEST_PACKAGES
+	@for pkg in $(SEQUENTIAL_TEST_PACKAGES); do \
+    	echo "Running tests for $$pkg..."; \
+    	go test -cover $$pkg || exit 1; \
+	done
```
otherview, line 66: I noticed that we're running tests on the API fees package sequentially.
Could you share the reasoning behind this approach?

freemanzMrojo: Yes, the reasoning is that the ticker signals are broadcasted sequentially but  (sometimes) not received in the same order (in the context of a "normal" execution does not matter because it ticks per every block). I am talking about https://github.com/vechain/thor/blob/master/co/signal.go.

This produces some flakiness in the tests when it happens. I have a sample execution [in this Github Action](https://github.com/vechain/thor/actions/runs/13078435874/job/36496210807). In case you want to follow the sequence of events I added some logs with the preffix `LLEGA` to track broadcast, arrival and cache status (especially, check the second occurrence of `LLEGA SetBestBlockID github.com/vechain/thor/v2/test/testchain.(*Chain).MintBlock and number 5` where the tick for block 5 does not arrive on time).

api/fees/fees.go
```diff
@@ -0,0 +1,239 @@
...
+			return
+		}
+	}
+}
+
+func (f *Fees) Close() {
+	close(f.done)
+}
+
+func (f *Fees) CacheLen() int {
```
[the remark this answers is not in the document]
freemanzMrojo: You are right, they were very useful over the development but I should say them goodbye indeed. Thanks!

api/doc/thor.yaml
```diff
@@ -752,6 +755,38 @@ paths:
...
+  /fees/history:
+    get:
+      tags:
+        - Fees
+      summary: Retrieve historical fee data
+      description: |
+        Retrieve historical fee data, including the gas used ratio per block.
+      parameters:
+        - $ref: '#/components/parameters/BlockCountInQuery'
+        - $ref: '#/components/parameters/NewestBlockInQuery'
```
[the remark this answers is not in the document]
freemanzMrojo: Unfortunately in Swagger, afaik, if you change the name of the parameter (`newestBlock` in this case), you cannot just reuse the existing reference (`RevisionInQuery`) even though they share the rest.

[In the implementation in Eth `newestBlock` is the name](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_feehistory/) + `revision` is misleading in this context IMO (otherwise the parameters would be `blockCount` and `revision`, so it is unclear whether we go up or down (should always be downwards).

api/utils/http.go
```diff
@@ -38,6 +38,22 @@ func BadRequest(cause error) error {
...
+	}
+}
+
+// NotFound convenience method to create http not found error.
+func NotFound(cause error) error {
+	return &httpError{
+		cause:  cause,
+		status: http.StatusNotFound,
+	}
+}
```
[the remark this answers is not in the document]
freemanzMrojo: In the other APIs it is always bad request for validation, regardless the parameter value (so a valid block in `blocks` that does not exist yet will return 400).

404 IMO is the more appropriate value in those cases.

api/fees/cache.go
```diff
@@ -0,0 +1,109 @@
...
+
+func (fc *FeesCache) get(blockID thor.Bytes32) (*FeeCacheEntry, uint32, error) {
+	if feeCacheEntry, blockNumberFloat64, ok := fc.cache.Get(blockID); ok {
+		return feeCacheEntry.(*FeeCacheEntry), uint32(blockNumberFloat64), nil
+	}
+
+	return nil, 0, errors.New("blocks fees not found")
+}
+
+func (fc *FeesCache) getByRevision(rev *utils.Revision, bft bft.Committer) (*FeeCacheEntry, uint32, error) {
```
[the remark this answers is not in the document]
freemanzMrojo: You mean avoided and have its logic somewhere else? Like just copy the lines and put them in the validation method?

api/fees/cache.go
```diff
@@ -0,0 +1,109 @@
...
+		return nil, 0, err
+	}
+	return fc.get(id)
+}
+
+func (fc *FeesCache) push(header *block.Header) {
+	fc.cache.Set(header.ID(), newFeesCacheEntry(header), float64(header.Number()))
+}
+
+func (fc *FeesCache) resolveRange(oldestBlockID thor.Bytes32, lastBlockNumber uint32) (uint32, []*hexutil.Big, []float64, bool, error) {
```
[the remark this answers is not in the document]
freemanzMrojo: This is the code where `resolveRange` is called so the validations are done at that stage: https://github.com/vechain/thor/blob/miguel/feehistory2/api/fees/fees.go#L163

We are doing something similar to what your code suggests, it is just that we need to know what are the limits since there will be times where we do not read the entire cache, that is why we have those 2 options (oldest block in cache, yes or no).

The code might look more complicated because of `prio_cache`'s `ForEach`, but looking at its implementation it looks like the way to go when it comes to iterating its entries.

freemanzMrojo:
> Looks pretty good!
> 
> I think it might be possible to simplify the overall logic. Perhaps something that makes responsibilities more decoupled, like :
> 
> * Validate and sanitize inputs
> * Request data from cache ( assumes requests have previously been validated )
> * If the cache doesn't have data, it backfills from the repo

Thanks!

This is exactly what the code does. A good example to see these 3 elements is the main method of the API: https://github.com/vechain/thor/blob/miguel/feehistory2/api/fees/fees.go#L163

Could you clarify what you mean by "decoupling"? Having these 3 elements in different files?

api/utils/revisions.go
```diff
@@ -70,17 +70,20 @@ func ParseRevision(revision string, allowNext bool) (*Revision, error) {
 	return &Revision{uint32(n)}, err
 }
 
-// GetSummary returns the block summary for the given revision,
-// revision required to be a deterministic block other than "next".
-func GetSummary(rev *Revision, repo *chain.Repository, bft bft.Committer) (sum *chain.BlockSummary, err error) {
+func ParseNumberRevision(revision uint32) *Revision {
+	return &Revision{revision}
+}
```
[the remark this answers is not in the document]
freemanzMrojo: Done, `NewRevision` since it is exposed through the `utils` package 👍

api/utils/revisions.go
freemanzMrojo: I am not sure, this logic was internal to `GetSummary`. Since the newest block will be in the new cache, the revision has to be parsed somehow.

api/doc/thor.yaml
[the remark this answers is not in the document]
otherview: NewestBlock makes sense 👍

api/api.go
otherview: Ah, yes yes, missed that 👍

api/utils/revisions.go
freemanzMrojo: This is done now, back to the previous version

api/fees/cache.go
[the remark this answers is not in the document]
freemanzMrojo: Deleted too

api/fees/fees.go
```diff
@@ -0,0 +1,239 @@
...
+	// Newest block should always be in the cache
+	_, newestBlockNumber, err := f.cache.getByRevision(newestBlock, f.bft)
+	if err != nil {
+		return 0, 0, utils.NotFound(errors.WithMessage(err, "newestBlock"))
+	}
+
+	return uint32(blockCount), newestBlockNumber, nil
+}
+
+func getOldestBlockNumber(blockCount uint32, newestBlock uint32) uint32 {
```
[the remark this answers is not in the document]
freemanzMrojo: Related to the explanation in today's catch-up, Ethereum does not return that request as an error, it just returns the values available.

api/fees/fees.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return utils.BadRequest(err)
+	}
+
+	cacheOldestBlockNumber, cacheBaseFees, cacheGasUsedRatios, shouldGetBlockSummaries, err := f.cache.resolveRange(oldestBlockID, newestBlockNumber)
+	if err != nil {
+		return utils.HTTPError(err, http.StatusInternalServerError)
+	}
+
+	if shouldGetBlockSummaries {
```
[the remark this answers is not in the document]
freemanzMrojo: I think I am going to resolve this comment in favour of the one with the code snippet. I think it is clear what you mean after today's conversation.

api/fees/fees.go
```diff
@@ -0,0 +1,239 @@
...
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
+	if err != nil {
+		return 0, 0, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	// Newest block should always be in the cache
+	_, newestBlockNumber, err := f.cache.getByRevision(newestBlock, f.bft)
+	if err != nil {
+		return 0, 0, utils.NotFound(errors.WithMessage(err, "newestBlock"))
+	}
```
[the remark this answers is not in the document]
freemanzMrojo: Also discussed in today's catch-up, there is a new version

api/fees/fees.go
```diff
@@ -0,0 +1,239 @@
...
+func getOldestBlockNumber(blockCount uint32, newestBlock uint32) uint32 {
+	oldestBlockInt32 := int32(newestBlock) + 1 - int32(blockCount)
+	oldestBlock := uint32(0)
+	if oldestBlockInt32 >= 0 {
+		oldestBlock = uint32(oldestBlockInt32)
+	}
+	return oldestBlock
+}
+
+func (f *Fees) getOldestBlockIDByNumber(oldestBlock uint32) (thor.Bytes32, error) {
```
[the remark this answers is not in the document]
freemanzMrojo: It is checked in `validateGetFeesHistoryParams`

api/utils/http.go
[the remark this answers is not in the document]
otherview: > In the other APIs it is always bad request for validation, regardless the parameter value (so a valid block in blocks that does not exist yet will return 400).

Is this correct ? I think requesting a non-existing block or transaction returns a 200 nil.

api/fees/cache.go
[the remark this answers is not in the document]
otherview: Discussed offline 👍

api/utils/revisions.go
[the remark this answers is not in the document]
otherview: That's fair. However, if you feels strongly about it, that might be done in a different PR which adds this method. There are some places that use `NewRevision(..)` and others `&Revision{revision}`.

api/utils/http.go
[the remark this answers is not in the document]
freemanzMrojo: I think you are right, this is the closest example that I have found https://github.com/vechain/thor/blob/master/api/blocks/blocks.go#L56

Do you want the same behaviour here?

api/fees/data.go
```diff
@@ -0,0 +1,213 @@
...
+			blockFee.blockSummary, blockFee.err = utils.GetSummary(blockFee.blockRevision, fd.repo, fd.bft)
+			if blockFee.blockSummary == nil {
+				blockFee.err = fmt.Errorf("block summary is nil for block number %d", blockNumber)
+			}
+		}
+		blockDataChan <- blockFee
+	}
+}
+
+func (fd *FeesData) processBlockSummaryRange(blockCount uint32, summary *chain.BlockSummary) (uint32, chan *blockData) {
```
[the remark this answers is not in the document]
freemanzMrojo: If what you want is something sequential I prefer to just implement it rather than having any benchmarks at this stage, since that means having 2 implementations. 

I really think that the current implementation is better. But I prefer to get this PR done tbh, so if you want something sequential please confirm it and then I will do it so it is approved.

api/fees/fees.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	// Too old
+	bestBlockSummary := f.data.repo.BestBlockSummary()
+	if bestBlockSummary == nil {
+		return 0, nil, nil, utils.HTTPError(errors.New("best block not found"), http.StatusInternalServerError)
+	}
+	newestBlockNumberSupported := getOldestBlockNumber(uint32(f.data.size), bestBlockSummary.Header.Number())
+	if newestBlockNumberSupported > newestBlockSummary.Header.Number() {
+		return 0, nil, nil, utils.BadRequest(errors.New(fmt.Sprintf("newestBlock must be between %d and %d", newestBlockNumberSupported, bestBlockSummary.Header.Number())))
+	}
```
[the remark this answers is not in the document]
freemanzMrojo: This condition validates that the `newestBlock` is in the range of the cache (that was the premise that I understood from the catch-up with Darren), that is why it is `size` instead of `maxBlocks`.

What is the second part of the or checking that is not checked already?

api/fees/fees.go
```diff
@@ -0,0 +1,140 @@
...
+	//blockCount validation
+	blockCountParam := req.URL.Query().Get("blockCount")
+	blockCountUInt64, err := strconv.ParseUint(blockCountParam, 10, 32)
+	if err != nil {
+		return 0, nil, nil, utils.BadRequest(errors.WithMessage(err, "invalid blockCount, it should represent an integer"))
+	}
+	blockCount := uint32(blockCountUInt64)
+	maxBlocks := uint32(math.Max(float64(f.data.backtraceLimit), float64(f.data.size)))
+	if blockCount < 1 || blockCount > maxBlocks {
+		return 0, nil, nil, utils.BadRequest(errors.New(fmt.Sprintf("blockCount must be between 1 and %d", f.data.size)))
```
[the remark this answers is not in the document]
otherview: Actually this should be `f.data.backtraceLimit`. As the maximum number of blocks should be the backtrace limit set on the node flag. Just realized it now.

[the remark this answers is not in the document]
freemanzMrojo: Not really, it depends. `maxBlocks` is actually `f.data.backtraceLimit` if bigger than the cache size.

api/fees/fees.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		if f.data.repo.IsNotFound(err) {
+			return 0, nil, nil, utils.NotFound(errors.WithMessage(err, "newestBlock"))
+		}
+		return 0, nil, nil, err
+	}
+	// Too old
+	bestBlockSummary := f.data.repo.BestBlockSummary()
+	if bestBlockSummary == nil {
+		return 0, nil, nil, utils.HTTPError(errors.New("best block not found"), http.StatusInternalServerError)
```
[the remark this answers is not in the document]
freemanzMrojo: Removed

api/fees/data.go
```diff
@@ -0,0 +1,213 @@
...
+		baseFee:      (*hexutil.Big)(baseFee),
+		gasUsedRatio: float64(header.GasUsed()) / float64(header.GasLimit()),
+	}
+}
+
+func (fd *FeesData) pushToCache(header *block.Header) {
+	fd.cache.Set(header.ID(), newFeesCacheEntry(header), float64(header.Number()))
+}
+
+func (fd *FeesData) resolveRange(oldestBlockSummary *chain.BlockSummary, newestBlockSummary *chain.BlockSummary, blockCount uint32) (uint32, []*hexutil.Big, []float64, error) {
```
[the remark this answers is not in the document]
freemanzMrojo: changed in a similar way

api/utils/http.go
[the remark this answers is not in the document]
freemanzMrojo: Done this way.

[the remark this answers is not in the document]
freemanzMrojo: Returning 200 nil now.

api/utils/http.go
```diff
@@ -38,6 +38,14 @@ func BadRequest(cause error) error {
 	}
 }
 
+// Forbidden convenience method to create http forbidden error.
```
otherview, line 41: nit: No need to change this 👍

api/utils/revisions.go
```diff
@@ -28,6 +28,10 @@ type Revision struct {
 	val interface{}
 }
 
+func NewRevision(revision uint32) *Revision {
```
otherview, line 31: Nit: Don't think this is used

api/fees/data.go
```diff
@@ -0,0 +1,101 @@
...
+	cacheSize := int(math.Min(float64(backtraceLimit), float64(fixedSize)))
+	maxBacktraceLimit := uint32(math.Max(float64(backtraceLimit), float64(cacheSize)))
+	return &FeesData{
+		repo:              repo,
+		cache:             cache.NewPrioCache(cacheSize),
+		bft:               bft,
+		cacheSize:         uint32(cacheSize),
+		maxBacktraceLimit: maxBacktraceLimit,
+	}
+}
```
otherview, line 31: It seems like the cache is competing with the backtrace limit.
It's fine to have fixed cache size.

```go
func newFeesData(repo *chain.Repository, bft bft.Committer, backtraceLimit uint32, fixedSize uint32) *FeesData {
	return &FeesData{
		repo:              repo,
		cache:             cache.NewPrioCache(int(fixedSize)),
		bft:               bft,
		maxBacktraceLimit: backtraceLimit,
	}
}
```

api/fees/data.go
```diff
@@ -0,0 +1,101 @@
...
+func newFeesCacheEntry(header *block.Header) *FeeCacheEntry {
+	baseFee := header.BaseFee()
+	if baseFee == nil {
+		baseFee = big.NewInt(0)
+	}
+	return &FeeCacheEntry{
+		baseFee:      (*hexutil.Big)(baseFee),
+		gasUsedRatio: float64(header.GasUsed()) / float64(header.GasLimit()),
+	}
+}
```
otherview, line 42: Only one usage, we can leverage the complexity to a different method. Perhaps getBaseFee ?
```suggestion
func getBaseFee(baseFee *big.Int) *hexutil.Big {
	if baseFee != nil {
		return (*hexutil.Big)(baseFee)
	}
	return (*hexutil.Big)(big.NewInt(0))
}

```

api/fees/data.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (fd *FeesData) pushToCache(header *block.Header) {
+	fd.cache.Set(header.ID(), newFeesCacheEntry(header), float64(header.Number()))
+}
```
otherview, line 46: nit: put the complexity in the entry constructor
```suggestion
func (fd *FeesData) pushToCache(header *block.Header) {
	feeCacheEntry := &FeeCacheEntry{
		baseFee:      getBaseFee(header.BaseFee()),
		gasUsedRatio: float64(header.GasUsed()) / float64(header.GasLimit()),
	}
	fd.cache.Set(header.ID(), feeCacheEntry, float64(header.Number()))
}

```

api/fees/data.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (fd *FeesData) resolveRange(newestBlockSummary *chain.BlockSummary, blockCount uint32) (uint32, []*hexutil.Big, []float64, error) {
```
otherview, line 48: It looks to me like there might be a few simplifications that could be done here.
I've started tweaking it to see if it would work, and ended up rewriting the method.
I didn't run the tests, so it might have misses, but I want to leave it here and when you're back available we can jump on a call and discuss it.

```
func (fd *FeesData) resolveRange(newestBlockSummary *chain.BlockSummary, blockCount uint32) (uint32, []*hexutil.Big, []float64, error) {
	// assumed these are always valid ranges
	newestBlockNumber := newestBlockSummary.Header.Number()
	oldestBlockNumber := newestBlockNumber - blockCount + 1

	// fetch entries from cache
	entries := make([]*cache.PrioEntry, blockCount)
	fd.cache.ForEach(func(ent *cache.PrioEntry) bool {
		if ent.Priority >= float64(oldestBlockNumber) && ent.Priority <= float64(newestBlockNumber) {
			entries[uint32(ent.Priority)-oldestBlockNumber] = ent
		}
		return true
	})

	// return calculated fees and ratios
	baseFees := make([]*hexutil.Big, blockCount)
	gasUsedRatios := make([]float64, blockCount)

	for i, ent := range entries {
		if ent == nil { // value not in cache
			// retrieve from db + retro-populate cache
			blockSummary, err := fd.repo.NewBestChain().GetBlockSummary(oldestBlockNumber + uint32(i))
			if err != nil {
				return 0, nil, nil, err
			}
			fd.pushToCache(blockSummary.Header)

			baseFees[i] = getBaseFee(blockSummary.Header.BaseFee())
			gasUsedRatios[i] = float64(blockSummary.Header.GasUsed()) / float64(blockSummary.Header.GasLimit())
			continue
		}

		// use cached values
		baseFees[i] = getBaseFee((*big.Int)(ent.Value.(*FeeCacheEntry).baseFee))
		gasUsedRatios[i] = ent.Value.(*FeeCacheEntry).gasUsedRatio
	}

	return oldestBlockNumber, baseFees, gasUsedRatios, nil
}
```

api/fees/fees.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+func New(repo *chain.Repository, bft bft.Committer, backtraceLimit uint32, fixedCacheSize uint32) *Fees {
+	return &Fees{
+		data: newFeesData(repo, bft, backtraceLimit, fixedCacheSize),
+		done: make(chan struct{}),
+	}
+}
+
+func getOldestBlockNumber(blockCount uint32, newestBlock uint32) uint32 {
```
otherview, line 27: nit: don't think this is needed

api/fees/fees.go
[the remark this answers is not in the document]
otherview: Addressed below, added a suggestion of the verification / fast-return

api/doc/thor.yaml
```diff
@@ -752,6 +755,38 @@ paths:
                 type: string
                 example: 'Invalid address'
 
+  /fees/history:
+    get:
+      tags:
+        - Fees
+      summary: Retrieve historical fee data
+      description: |
+        Retrieve historical fee data, including the gas used ratio per block.
```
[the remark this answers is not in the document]
freemanzMrojo: Done

api/fees/fees.go
```diff
@@ -0,0 +1,121 @@
...
+	//blockCount validation
+	blockCountParam := req.URL.Query().Get("blockCount")
+	blockCountUInt64, err := strconv.ParseUint(blockCountParam, 10, 32)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "invalid blockCount, it should represent an integer"))
+	}
+	blockCount := uint32(blockCountUInt64)
+	if blockCount < 1 || blockCount > f.data.maxBacktraceLimit {
+		return 0, nil, utils.BadRequest(errors.New(fmt.Sprintf("blockCount must be between 1 and %d", f.data.maxBacktraceLimit)))
+	}
```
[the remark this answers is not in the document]
freemanzMrojo: Done

api/fees/data.go
freemanzMrojo: Done

api/fees/data.go
freemanzMrojo: Done

api/fees/data.go
freemanzMrojo: Done

api/fees/fees.go
freemanzMrojo: Done

api/fees/data.go
freemanzMrojo: Done

api/utils/http.go
freemanzMrojo: Done

api/fees/fees.go
```diff
@@ -0,0 +1,121 @@
...
+			f.data.pushToCache(bestBlockSummary.Header)
+		case <-f.done:
+			return
+		}
+	}
+}
+
+func (f *Fees) Close() {
+	close(f.done)
+}
```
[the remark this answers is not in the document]
freemanzMrojo: Done

api/utils/revisions.go
freemanzMrojo: Done

api/fees/types.go
```diff
@@ -0,0 +1,36 @@
...
+	baseFee      *hexutil.Big
+	gasUsedRatio float64
+}
+type FeesData struct {
+	repo           *chain.Repository
+	cache          *cache.PrioCache
+	bft            bft.Committer
+	backtraceLimit uint32 // The max number of blocks to backtrace.
+}
+type GetFeesHistory struct {
```
libotony, line 32:
```suggestion
type FeesHistory struct {
```

Better?

api/fees/types.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/vechain/thor/v2/bft"
+	"github.com/vechain/thor/v2/cache"
+	"github.com/vechain/thor/v2/chain"
+)
+
+type Fees struct {
```
libotony, line 17: Better move `Fees` into `fees.go` and same with `FeesData`, `FeeCacheEntry`.

api/doc/thor.yaml
```diff
@@ -2259,6 +2327,13 @@ components:
       schema:
         type: string
 
+    NewestBlockInQuery:
```
libotony, line 2330: This seems to be the same concept as `revision`, wdyt?

libotony: Coming back to this since I found that the API in ethereum only support block number and tags but node block ID. I would think it's not necessary to work with block ID in this API.

freemanzMrojo: you mean a type that is essentially `revision` without blockID? so:

```
func ParseRevision(revision string, allowNext bool) (*Revision, error) {
	if revision == "" || revision == "best" {
		return &Revision{revBest}, nil
	}

...
        // WITHOUT THIS
	if len(revision) == 66 || len(revision) == 64 {
		blockID, err := thor.ParseBytes32(revision)
		if err != nil {
			return nil, err
		}
		return &Revision{blockID}, nil
	}

...
	return &Revision{uint32(n)}, err
}
```

api/fees/types.go
freemanzMrojo: done

api/doc/thor.yaml
freemanzMrojo: FYI `newestBlock` is indeed a `revision`, the name change is:

1. Because having `blockCount` and `revision` is confusing (going upwards or downwards?)
2. Because `newestBlock` is the name in Ethereum

libotony: > you mean a type that is essentially revision without blockID? 
Yes, just think of the necessity of having a blockID as the argument.

freemanzMrojo: Ok, since the name of the new parser might be opinionated, and `ParseNewestBlock` does not seem generic enough, do you have any suggestions? The logic should be reused by the current `ParseRevision` IMO since they share common code.

To sum up, any suggestions abut the method name to parse this?

freemanzMrojo: Called it for now `ParseRevisionWithoutBlockID`, thanks

libotony:
Correct me if I'm not describing correctly, current implementation is a `actively update` model. Which is not able two handle the situation when there is a fork, to have more than one block branch(BlockID seems not used).

I would suggest consider changing to the model to a `pull based` model, which 

1. Resolve `OldestBlock` to a blockID
2. Get blockID from cache, if not exist fetch from repository
3. Caching the result with parentID for fast iterating

Moreover, coming back to the topic regarding `revision`, I would suggest using `revison` to replace `OldestBlock`, reasons:

1.  Revision accepts blockID, with blockID the response are deterministic, which will be preferred for any fork-aware clients
2. Aligns with all our API specs
3. Ethereum doesn't have temporary fork as frequent as we do

@freemanzMrojo @paologalligit

freemanzMrojo:
> Correct me if I'm not describing correctly, current implementation is a `actively update` model. Which is not able two handle the situation when there is a fork, to have more than one block branch(BlockID seems not used).
> 
> I would suggest consider changing to the model to a `pull based` model, which
> 
> 1. Resolve `OldestBlock` to a blockID
> 2. Get blockID from cache, if not exist fetch from repository, then compute the `FeeCacheEntry`
> 3. Caching the result with parentID for fast iterating in `FeeCacheEntry`
> 
> Moreover, coming back to the topic regarding `revision`, I would suggest using `revison` to replace `OldestBlock`, reasons:
> 
> 1. Revision accepts blockID, with blockID the response are deterministic, which will be preferred for any fork-aware clients
> 2. Aligns with all our API specs
> 3. Ethereum doesn't have temporary fork as frequent as we do, so only block number works with ethereum but we'd better work with blockIDs
> 
> @freemanzMrojo @paologalligit

Hi @libotony , I implemented a first version in the last commit, let me know what you think and please comment inline that new method `computeOldestBlockRevision` (I had some doubts when dealing with what you called "caching the result with parentID"). Thanks!

Makefile
freemanzMrojo: Tick not used anymore so test execution back to previous config

freemanzMrojo:
@libotony just implemented the changes according to our last conversation, relying just on block IDs + cache pusher on tick method removed.

please have a review when you have a chance, thanks!

api/fees/data.go
```diff
@@ -0,0 +1,81 @@
...
+func (fd *FeesData) pushToCache(header *block.Header) {
+	feeCacheEntry := &FeeCacheEntry{
+		baseFee:       getBaseFee(header.BaseFee()),
+		gasUsedRatio:  float64(header.GasUsed()) / float64(header.GasLimit()),
+		parentBlockID: header.ParentID(),
+	}
+	fd.cache.Set(header.ID(), feeCacheEntry, float64(header.Number()))
+}
+
+func (fd *FeesData) resolveRange(newestBlockSummary *chain.BlockSummary, blockCount uint32) (*utils.Revision, []*hexutil.Big, []float64, error) {
```
otherview: Worth adding a comment that it's expected that the ranges of newestBlockSummary - blockCount are already sanitized.

otherview (commented):
Yeah, that's really good stuff.
Added two minor nits, other than that lgtm 👍

api/fees/data.go
otherview: Not sure why that comment was lost but I think it's ok to replace
```
newestBlockID, err := fd.repo.NewBestChain().GetBlockID(newestBlockSummary.Header.Number())
	if err != nil {
		return thor.Bytes32{}, nil, nil, err
```

with 
`newestBlockID := newestBlockSummary.Header.ID()`

Makefile
```diff
@@ -44,16 +46,46 @@ go_version_check:
...
 $(CURDIR)/bin/thor \
 $(CURDIR)/bin/disco
 
+clean: #@ Clean the test and build cache and remove binaries
+	@echo "cleaning test cache..."
+	@go clean -testcache
+	@echo "cleaning build cache and binaries..."
+	@go clean -cache -modcache -i -r
+	@rm -rf $(CURDIR)/bin/*
+	@echo "done. build cache and binaries removed."
```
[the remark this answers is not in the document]
freemanzMrojo: Maybe as part of a different part targeting `master` 👍

api/fees/fees.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/vechain/thor/v2/bft"
+	"github.com/vechain/thor/v2/cache"
+	"github.com/vechain/thor/v2/chain"
+	"github.com/vechain/thor/v2/thor"
+)
+
+type Fees struct {
+	data *FeesData
+}
+type FeeCacheEntry struct {
```
libotony, line 27: It might be not clear for my previous comment. I was suggesting move `FeeCacheEntry` and `FeesData` to `data.go`. Reason behind that is when I was trying to find the type define near function `New`, put the constructor and struct definition will make the file structure clear, wdyt?

api/fees/fees.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	// Validate newestBlock
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	newestBlockSummary, err := utils.GetSummary(newestBlock, f.data.repo, f.data.bft)
+	if err != nil {
```
libotony, line 65: Suggest returning bad request if `blockCount` inputs 0

api/fees/fees.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/vechain/thor/v2/api/utils"
+	"github.com/vechain/thor/v2/bft"
+	"github.com/vechain/thor/v2/cache"
+	"github.com/vechain/thor/v2/chain"
+	"github.com/vechain/thor/v2/thor"
+)
+
+type Fees struct {
+	data *FeesData
+}
```
libotony, line 26:
```suggestion
type Fees struct {
	bft    bft.Committer
	backtraceLimit uint32
	data *FeesData
}
```

Needs to be formatted

api/fees/fees.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Calculate minAllowedBlock
+	minAllowedBlock := uint32(math.Max(0, float64(int(bestBlockSummary.Header.Number())-int(f.data.backtraceLimit)+1)))
+
+	// Adjust blockCount if necessary
+	if int(bestBlockSummary.Header.Number()) < int(f.data.backtraceLimit) {
+		blockCount = uint64(math.Min(float64(blockCount), float64(bestBlockSummary.Header.Number()+1)))
+	}
+
+	if newestBlockSummary.Header.Number() < minAllowedBlock {
```
libotony, line 77: Suggest returning bad request error, treat them as user input error

freemanzMrojo: Done

api/fees/fees.go
```diff
@@ -0,0 +1,117 @@
...
+
+	// Validate newestBlock
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	newestBlockSummary, err := utils.GetSummary(newestBlock, f.data.repo, f.data.bft)
+	if err != nil {
+		return 0, nil, err
```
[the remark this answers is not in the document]
freemanzMrojo: Done

api/fees/fees.go
freemanzMrojo: Done

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+	// Validate newestBlock
+	newestBlock, err := utils.ParseRevision(req.URL.Query().Get("newestBlock"), false)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	newestBlockSummary, err := utils.GetSummary(newestBlock, f.data.repo, f.bft)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
```
libotony, line 55:
```suggestion
	if err != nil {
		if a.repo.IsNotFound(err) {
		        // return 400 for the parameter validation
			return utils.BadRequest(errors.WithMessage(err, "newestBlock"))
		}
		// all other unexpected errors will fall to 500 error
		return err
	}
```
If it's not found return a bad request, other errors are not expected, return internal server error instead ref https://github.com/vechain/thor/blob/master/api/utils/http.go#L76-L85

libotony: @otherview Please do another verify.

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+
+	bestBlockSummary := f.data.repo.BestBlockSummary()
+	// Calculate minAllowedBlock
+	minAllowedBlock := uint32(math.Max(0, float64(int(bestBlockSummary.Header.Number())-int(f.backtraceLimit)+1)))
```
libotony, line 59: Could you explain a bit regarding the +1 here? How do you define the range? Given best block to be 8000(back trace limit 1000), what's the valid range of allowed block does [7000, 800] sound valid to you?

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	newestBlockSummary, err := utils.GetSummary(newestBlock, f.data.repo, f.bft)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "newestBlock"))
+	}
+
+	bestBlockSummary := f.data.repo.BestBlockSummary()
```
libotony, line 57:
```suggestion
	bestBlockNum := f.data.repo.BestBlockSummary().Header().Number()
```

Minor, but trying to make following code clear to read.

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/vechain/thor/v2/chain"
+)
+
+type Fees struct {
+	data           *FeesData
+	bft            bft.Committer
+	backtraceLimit uint32 // The max number of blocks to backtrace.
+}
+
+func New(repo *chain.Repository, bft bft.Committer, backtraceLimit uint32, fixedCacheSize uint32) *Fees {
```
libotony, line 26: Would you consider make `fixedCacheSize` a internal constant?

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (f *Fees) validateGetFeesHistoryParams(req *http.Request) (uint32, *chain.BlockSummary, error) {
+	// Validate blockCount
+	blockCountParam := req.URL.Query().Get("blockCount")
+	blockCount, err := strconv.ParseUint(blockCountParam, 10, 32)
+	if err != nil {
+		return 0, nil, utils.BadRequest(errors.WithMessage(err, "invalid blockCount, it should represent an integer"))
+	}
+
+	if blockCount == 0 {
```
libotony, line 42: Given the conversation we had offline, do you think it's necessary to also have the check of `blockCount>f.backtraceLimit {return utils.BadRequest}` ? @otherview

freemanzMrojo: No because in that case we just return what is available

api/fees/data.go
```diff
@@ -0,0 +1,86 @@
...
+func (fd *FeesData) resolveRange(newestBlockSummary *chain.BlockSummary, blockCount uint32) (thor.Bytes32, []*hexutil.Big, []float64, error) {
+	newestBlockID := newestBlockSummary.Header.ID()
+
+	baseFees := make([]*hexutil.Big, blockCount)
+	gasUsedRatios := make([]float64, blockCount)
+
+	var oldestBlockID thor.Bytes32
+	for i := blockCount; i > 0; i-- {
+		oldestBlockID = newestBlockID
+		fees, _, found := fd.cache.Get(newestBlockID)
```
libotony, line 62: how about a function called `fd.getFeeEentry(blockID thor.bytes32) (*FeeCacheEntry, error)` to warp up a `get or load` logic?

api/fees/data.go
```diff
@@ -0,0 +1,86 @@
...
+			blockSummary, err := fd.repo.GetBlockSummary(newestBlockID)
+			if err != nil {
+				return thor.Bytes32{}, nil, nil, err
+			}
+
+			fd.pushToCache(blockSummary.Header)
+
+			baseFees[i-1] = getBaseFee(blockSummary.Header.BaseFee())
+			gasUsedRatios[i-1] = float64(blockSummary.Header.GasUsed()) / float64(blockSummary.Header.GasLimit())
+
```
libotony, line 74: A minor safety guard
``` go
if block.Number(newestBlockID) == 0{
        break
}
```

libotony: Do you think it's good to have?

api/fees/fees.go
freemanzMrojo: `minAllowedBlock` is used as a number of blocks (so it is kind of the "same family" as `blockCount`). That is the reason for + 1. You can check the usage below.

With `blockCount`, we will have 10 blocks for a range between 0 and 9. So `newestBlock` (`best`) would be 9 and blockCount >= 10 will return the 10 blocks.

api/fees/data.go
freemanzMrojo: This is already controlled by the validation of `blockCount` in the previous stage so it should not be necessary.

api/fees/fees.go
freemanzMrojo: Just FYI all these validation have been simplified several times to the current state, and have been tested extensively, not only by the integration tests of this PR but also using the e2e test repo (PR in there to be created).

Feel free to suggest more integration tests if you think we should check an edge case 👍

api/fees/fees.go
freemanzMrojo: That would make the tests break because I would not be able to control the cache size to test this (right now we use 10 blocks, so if having 1024 every I would have issues to validate scenarios)

freemanzMrojo: I might create a constant at `api.go`, that would work in any case, please confirm that is ok for you

api/fees/data.go
freemanzMrojo: I could do it, but, it is not already pretty simple and straightforward this method? I think it could harm readability the fact that we need to check another method to see what is happening.

api/fees/fees.go
freemanzMrojo: Added to `api.go`

api/fees/data.go
libotony: Ok, could you add a test case that best block number is lower than block count?

api/fees/fees.go
libotony: What do you mean by "available"?

api/fees/data.go
libotony: The computation has been run twice in `pushToCache`  and outside, which I would suggest being avoided.

api/fees/fees.go
```diff
@@ -0,0 +1,102 @@
...
+
+func (f *Fees) handleGetFeesHistory(w http.ResponseWriter, req *http.Request) error {
+	blockCount, newestBlockSummary, err := f.validateGetFeesHistoryParams(req)
+	if err != nil {
+		return err
+	}
+
+	oldestBlockRevision, baseFees, gasUsedRatios, err := f.data.resolveRange(newestBlockSummary, blockCount)
+	if err != nil {
+		return utils.HTTPError(err, http.StatusInternalServerError)
```
libotony, line 86:
```suggestion
		return err
```

Will be handled by `utils.WrapHandlerFunc`

api/fees/fees.go
freemanzMrojo: Following the other example, if we have 10 blocks, backtrace limit is 5 and we request from newestBlock 10 to blockCount 6, we return 5 blocks. That is how it works in Ethereum btw.

api/fees/data.go
freemanzMrojo: There is a test already here https://github.com/vechain/thor/blob/miguel/feehistory2/api/fees/fees_test.go#L239. We have 10 blocks overall and we request 11.

api/fees/fees.go
freemanzMrojo: Done

api/fees/data.go
freemanzMrojo: With "computation" you mean the assignment for `baseFees`, `gasUsedRatios` and so on?

freemanzMrojo: I can make `pushToCache` return the `FeeCacheEntry` (and assign it to `fees`) so the below computation does not happen and it is just 

```
		baseFees[i-1] = getBaseFee((*big.Int)(fees.(*FeeCacheEntry).baseFee))
		gasUsedRatios[i-1] = fees.(*FeeCacheEntry).gasUsedRatio

		newestBlockID = fees.(*FeeCacheEntry).parentBlockID
```

let me know what you think

api/fees/fees.go
otherview: Yep ✅

freemanzMrojo: Done

api/fees/fees.go
otherview: For context, the cut allows for a user to get blockCount == f.backtraceLimit easily. Otherwise it would be very easy to have an off by 1 request.

api/fees/data.go
libotony: Would you mind to put the logic of `pushToCache` into the code above?

api/fees/fees.go
libotony: Confirmed with @otherview, he is good with current suggestion.

api/fees/fees.go
libotony: I believe it is beneficial to alert users about obvious input errors. Notifying them about an 'out of range blockCount' can provide helpful guidance during the development stage.

api/fees/data.go
freemanzMrojo: Done

api/fees/fees.go
libotony: I was considering `blockCount>f.backtraceLimit` to be a user input error. And cut off should work if user input`blockID=100, blockCount=10, backtraceLimit=10` and when the request comes `bestBlock =101`, then cut off has value for user since the blockchain ticked right during the time of processing the block. @otherview @freemanzMrojo wdyt?

libotony (approved):

PR #987, opened 2025-02-25
BaseFee, maxFeePerGas and maxPriorityFeePerGas as hex

paologalligit:
# Description

This PR returns the `baseFee`, `maxFeePerGas`, `maxPriorityFeePerGas` as hex values in `/transactions` and `/blocks` api endpoints response.

[GHIssue](https://github.com/vechain/protocol-board-repo/issues/440)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

# Checklist:

### DIFF
--- a/api/blocks/types.go
+++ b/api/blocks/types.go
@@ -16,24 +16,24 @@ import (
 )
 
 type JSONBlockSummary struct {
-	Number       uint32       `json:"number"`
-	ID           thor.Bytes32 `json:"id"`
-	Size         uint32       `json:"size"`
-	ParentID     thor.Bytes32 `json:"parentID"`
-	Timestamp    uint64       `json:"timestamp"`
-	GasLimit     uint64       `json:"gasLimit"`
-	Beneficiary  thor.Address `json:"beneficiary"`
-	GasUsed      uint64       `json:"gasUsed"`
-	TotalScore   uint64       `json:"totalScore"`
-	TxsRoot      thor.Bytes32 `json:"txsRoot"`
-	TxsFeatures  uint32       `json:"txsFeatures"`
-	StateRoot    thor.Bytes32 `json:"stateRoot"`
-	ReceiptsRoot thor.Bytes32 `json:"receiptsRoot"`
-	COM          bool         `json:"com"`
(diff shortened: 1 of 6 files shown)


api/blocks/types.go
```diff
@@ -121,7 +121,7 @@ func buildJSONBlockSummary(summary *chain.BlockSummary, isTrunk bool, isFinalize
 		COM:          header.COM(),
 		IsTrunk:      isTrunk,
 		IsFinalized:  isFinalized,
-		BaseFee:      summary.Header.BaseFee(),
+		BaseFee:      (*math.HexOrDecimal256)(summary.Header.BaseFee()),
```
freemanzMrojo, line 124: Rather a question so we might decide within this PR. I can see that in geth they use `math.HexOrDecimal256` or `hexutil.Big` to represent base fee, max fee per gas and max priority fee per gas (depending on the place of the code).

However it feels confusing to me tbh. Do you think we should make a decision (so we use one or another) or just let it be like in geth (it is kind of a difficult decision because if a different dev jumps on a ticket and looks at geth, might use the option discarded by us because it is what geth does)?

Or maybe you understand why `hexutil.Big` is used in some places and this one in others.

freemanzMrojo: Actually just saw that we are already using both. Feel free to resolve conversation if you think that nothing should be done in this regard 👍

darrenvechain: I think for response bodies it doesn't really matter. I think its more important for client request bodies, if we accept hex and int.

`math.HexOrDecimal256` - both of these are valid: [1234, 0xabc1234]
`hexutil.Big` - I think only hex format is valid [0xabc1234]

freemanzMrojo: Saw that difference, the part that is confusing is that both types are used to represent the same values I mentioned in my first comment.

Anyways, it has been clarified already with @paologalligit offline. Thanks guys!

freemanzMrojo (approved):

darrenvechain (approved):

PR #988, opened 2025-02-26
feat: flag `min-effective-priority-fee` in transactions

freemanzMrojo:
# Description

New flag `--min-effective-priority-fee` that enforces the use of an effective priority fee (the minimum between the `maxPriorityFeePerGas` and `maxFeePerGas` minus the block base fee) greater than 0 in order to submit new transactions. 

Closes https://github.com/vechain/protocol-board-repo/issues/413


YeahNotSewerSide:
Hi, do you also need to propagate this flag and perform checks on transactions in validation in the [consensus](https://github.com/vechain/thor/blob/45ba5d6a70b261eb2929ca2b75bad4d0fbf16777/consensus/validator.go#L257)? Just wonder, since this check is present in the packer

freemanzMrojo:
> Hi, do you also need to propagate this flag and perform checks on transactions in validation in the [consensus](https://github.com/vechain/thor/blob/45ba5d6a70b261eb2929ca2b75bad4d0fbf16777/consensus/validator.go#L257)? Just wonder, since this check is present in the packer

Hi @YeahNotSewerSide, as far I see It wont be necessary because of this sequence of events:

```
	var goes co.Goes
	goes.Go(func() { n.comm.Sync(ctx, n.handleBlockStream) })
	goes.Go(func() { n.houseKeeping(ctx) })
	goes.Go(func() { n.txStashLoop(ctx) })
	goes.Go(func() { n.packerLoop(ctx) })

	goes.Wait()
```

The processing you are pointing out happens when calling `handleBlockStream` (so processing an existing block) while the current implementation is relying on `packerLoop` (packing a new block). As far as I understand, the transaction should not be included at the time `validate` is executed.

cmd/thor/main.go
```diff
@@ -302,6 +309,7 @@ func defaultAction(ctx *cli.Context) error {
 		p2pCommunicator.Communicator(),
 		ctx.Uint64(targetGasLimitFlag.Name),
 		skipLogs,
+		requireTxPriorityFee,
```
paologalligit, line 312: Do you think it's worth start grouping all the flags/options under one struct like `NodeOption` (something like the [mempool options](https://github.com/vechain/thor/blob/master/txpool/tx_pool.go#L55))? We could group together  
```
...
ctx.Uint64(targetGasLimitFlag.Name),
skipLogs,
requireTxPriorityFee,
...
```
I'm not saying that this needs to be done in this ticket, but maybe it can be done in another one. wdyt?

docs/usage.md
```diff
@@ -157,37 +157,38 @@ To show usages of all command line options:
...
+| `--cache`                           | Megabytes of RAM allocated to trie nodes cache (default: 4096)                                       |
+| `--disable-pruner`                  | Disable state pruner to keep all history                                                             |
+| `--enable-metrics`                  | Enables the metrics server                                                                           |
+| `--metrics-addr`                    | Metrics service listening address                                                                    |
+| `--enable-admin`                    | Enables the admin server                                                                             |
+| `--admin-addr`                      | Admin service listening address                                                                      |
+| `--txpool-limit-per-account`        | Transaction pool size limit per account                                                              |
+| `--require-max-priority-fee-per-gas`| Requires that all transactions have the maxPriorityFeePerGas attribute set                           |
```
claytonneal, line 189: if True it means field has to exist and be > 0
if False field can be null or non existant?
thanks

freemanzMrojo: Yes

cmd/thor/main.go
freemanzMrojo: I think it makes total sense, since I will be waiting for more reviewers I think I can just do it in the meantime here, thanks.

freemanzMrojo: Done now

packer/flow_test.go
```diff
@@ -364,3 +364,53 @@ func TestAdoptAfterGalacticaLowerBaseFeeThreshold(t *testing.T) {
 	best, _ := chain.BestBlock()
 	fmt.Println(best.Header().BaseFee())
 }
+
+func TestAdoptAfterGalacticaRequireMaxPriorityFee(t *testing.T) {
```
paologalligit, line 368: I would also add a legacy tx with `gasPriceCoef == 0` and expect it to success

paologalligit (commented):
LGTM :white_check_mark:

paologalligit (approved):

YeahNotSewerSide (approved):
lgtm

YeahNotSewerSide (approved):

paologalligit (approved):

packer/packer.go
```diff
@@ -52,7 +52,7 @@ func New(
 }
 
 // Schedule schedule a packing flow to pack new block upon given parent and clock time.
-func (p *Packer) Schedule(parent *chain.BlockSummary, nowTimestamp uint64) (flow *Flow, err error) {
+func (p *Packer) Schedule(parent *chain.BlockSummary, nowTimestamp uint64, minTxPriorityFee uint64) (flow *Flow, err error) {
```
libotony, line 55: `minTxPriorityFee` to be a property of `Packer` and can be access by flow via `f.packer.minTxPriorityFee`

packer/flow.go
```diff
@@ -89,6 +93,17 @@ func (f *Flow) hasTx(txid thor.Bytes32, txBlockRef uint32) (bool, error) {
 	return f.runtime.Chain().HasTransaction(txid, txBlockRef)
 }
 
+func (f *Flow) maxPriorityFeePerGasTooLow(t *tx.Transaction) error {
```
libotony, line 96:
```suggestion
func (f *Flow) validatePriorityFeePerGas(t *tx.Transaction) error {
```

cmd/thor/main.go
```diff
@@ -293,6 +295,17 @@ func defaultAction(ctx *cli.Context) error {
 		defer func() { log.Info("stopping pruner..."); pruner.Stop() }()
 	}
 
+	minTxPriorityFee := ctx.Uint64(minMaxPriorityFeePerGasFlag.Name)
+	if minTxPriorityFee > 0 {
+		log.Info("max priority fee per gas is required in transactions")
```
libotony, line 300: It's better to have value printed here.

cmd/thor/flags.go
```diff
@@ -183,6 +183,12 @@ var (
 		Usage: "comma separated list of allowed API tracers(none,all,call,prestate etc.)",
 	}
 
+	minMaxPriorityFeePerGasFlag = cli.Uint64Flag{
+		Name:  "minimum-max-priority-fee-per-gas",
```
libotony, line 187: Should be `minimum-priority-fee-per-gas`, we are verifying the 'acting' priority fee per gas here.

cmd/thor/flags.go
```diff
@@ -183,6 +183,12 @@ var (
...
+		Value: 0,
+		Usage: "minimum max priority fee per gas for all transactions",
```
libotony, line 189:
```suggestion
		Usage: "block proposer only flag, to set a minimum priority fee per gas for transactions to be included into block",
```

This description doesn't express that the value is checked via `>`, maybe we should set this flag's default value to 1?

cmd/thor/flags.go
freemanzMrojo: The name of the attribute in the transaction is `maxPriorityFeePerGas`, I think the current name makes more sense since that is what we are validating.

packer/flow.go
freemanzMrojo: I think the current name makes more sense since it is more specific. `validatePriorityFeePerGas` does not express that we are validating a `maxPriorityFeePerGas` value is >= than the set in the flag.

cmd/thor/flags.go
freemanzMrojo: I think 0 is the correct value for the following reasons:

1. If 1, we are essentially making the flag required since even though we do not set anything we will be requesting a `maxPriorityFeePerGas` >= 1 (so we wont be allowing 0s in practice)
2. If 1 is the minimum value, rather than a flag the value should be set at the attribute level
3. The description is accurate because the value set there is effectively the minimum, so >= 0. At `flow.go` we have the following:

```
	maxPriorityFeeTooLow := t.MaxPriorityFeePerGas() == nil || t.MaxPriorityFeePerGas().Cmp(f.minTxPriorityFee) < 0
	if flagIsSet && maxPriorityFeeTooLow {
		return badTxError{"max priority fee per gas too low"}
	}
```

So `t.MaxPriorityFeePerGas().Cmp(f.minTxPriorityFee) >= 0` (the opposite) would allow transactions to be adopted.

packer/packer.go
freemanzMrojo: `minTxPriorityFee` is only used by `flow.go` and its constructor `newFlow` is private and called from here. The rest of the `packer` does not use it so I would argue that is rather a `Flow` property.

cmd/thor/main.go
freemanzMrojo: Done

cmd/thor/flags.go
libotony: I think the feature should be validating effective priority fee per gas other than max priority fee per gas, since `effectivePriorityFeePerGas = min(tx.maxPriorityFeePerGas, tx.maxFeePerGas - block.baseFeePerGas)`

packer/flow.go
libotony: For function names, I think verbs is preferred instead of nouns, if you prefer `maxPriorityFeePerGasTooLow` then it should be `isMaxPriorityFeePerGasTooLow(t *tx.Transaction) bool`.

cmd/thor/flags.go
freemanzMrojo: If that is the case, does this comment still apply as is? https://github.com/vechain/protocol-board-repo/issues/413#issuecomment-2705738256

packer/packer.go
libotony: Yes, agree with you that it is Flow only variable. I am thinking the way to ease our life. By adding it to `Flow`'s parent `Packer`, we don't need to pass it every time the `Schedule` was called. Took a look into the code structure, I think it's better to be inside Packer and let Flow inherit that property.

packer/flow.go
freemanzMrojo: Ok, I can use `is...`. Won't change it just now until the staff related to the effective priority fee gets clarified, because it will affect the name of this method.

cmd/thor/flags.go
libotony: I see your point, sorry for the confusing. I am considering the minimum `effectivePriorityFeePerGas` should be 1 otherwise validators are doing works for free, would you agree?

packer/packer.go
freemanzMrojo: Well, if you put it in the `packer` you need to pass it to the constructor. [The original version of this code had it there actually and I changed it to this](https://github.com/vechain/thor/pull/988/commits/9d51934c004c218ef6b493d469cfd2cc4162c6cc). You can see that in terms of changes it is pretty much the same, so I do not think it makes things easier tbh.

cmd/thor/flags.go
freemanzMrojo: Done assuming this

cmd/thor/main.go
```diff
@@ -293,6 +295,17 @@ func defaultAction(ctx *cli.Context) error {
 		defer func() { log.Info("stopping pruner..."); pruner.Stop() }()
 	}
 
+	minTxPriorityFee := ctx.Uint64(minEffectivePriorityFeeFlag.Name)
+	if minTxPriorityFee > 0 {
+		log.Info(fmt.Sprintf("the minimum effective priority fee required in transactions is %d", minTxPriorityFee))
```
libotony, line 300:
```suggestion
		log.Info(fmt.Sprintf("the minimum effective priority fee required in transactions is %d wei", minTxPriorityFee))
```

packer/packer.go
libotony: I am more into your initial version as `Schedule` doesn't need to know about `effective priority fee`. Makes more sense to me when `effective priority fee` becomes `Packer`'s config property.

cmd/thor/flags.go
```diff
@@ -183,6 +183,12 @@ var (
 		Usage: "comma separated list of allowed API tracers(none,all,call,prestate etc.)",
 	}
 
+	minEffectivePriorityFeeFlag = cli.Uint64Flag{
+		Name:  "minimum-effective-priority-fee",
+		Value: 0,
+		Usage: "minimum effective priority fee for all transactions",
```
libotony, line 189:
```suggestion
		Name:  "min-effective-priority-fee",
		Value: 0,
		Usage: "set a minimum effective priority fee for transactions to be included in the block proposed by the block proposer",
```

cmd/thor/main.go
freemanzMrojo: Done

packer/packer.go
freemanzMrojo: Done

cmd/thor/flags.go
freemanzMrojo: Done

libotony (approved):

PR #992, opened 2025-03-03
feat: `fees/priority` flags

freemanzMrojo:
# Description

Added flags so we can find out the best config for priority fees without modifying the code.


api/fees/data.go
```diff
@@ -16,10 +16,13 @@ import (
 	"github.com/vechain/thor/v2/tx"
 )
 
-const (
-	priorityNumberOfTxsPerBlock = 3
-	priorityPercentile          = 60
-)
+type Config struct {
```
paologalligit, line 19: Do you think all these values can be turned into regular `int`? I see they are always cast to `int`

freemanzMrojo: Done, they are always positive but it is true that we save some casts and that it is also controlled by the flag itself.

paologalligit (approved):

PR #994, opened 2025-03-05
fix: use best block ID for txs if provided is higher number

darrenvechain:
# Description

This PR seeks to reduce the `leveldb not found` error when clients are requesting a tx receipt with a given block ID that is higher than the best. (this happens a lot with ALB + connex)

# Checklist:

### DIFF
--- a/api/transactions/transactions.go
+++ b/api/transactions/transactions.go
@@ -15,6 +15,7 @@ import (
 	"github.com/vechain/thor/v2/api/utils"
 	"github.com/vechain/thor/v2/chain"
 	"github.com/vechain/thor/v2/thor"
+	"github.com/vechain/thor/v2/tx"
 	"github.com/vechain/thor/v2/txpool"
 )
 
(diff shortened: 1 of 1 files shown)


api/transactions/transactions.go
```diff
@@ -210,6 +211,10 @@ func (t *Transactions) parseHead(head string) (thor.Bytes32, error) {
 	if err != nil {
 		return thor.Bytes32{}, err
 	}
+	// if the provided head is newer than the best block, return the best block ID.
```
libotony, line 214: This will have side effects in soft fork scenario, for example, we have block A(10) as best but somehow replaced block B(9) as best, then user sets A as head should request from a chain that using A as head. If user inputs an blockID then I would assume user have already seen the block, we shouldn't do the overwrite. The improvement can be done via

- Improve the error message
- If it's caused by chain state not synced for different node instance, this might not be able to be eliminated from node software itself.

PR #997, opened 2025-03-07
fix: add an error log for WriteJSON

darrenvechain:
# Description

If WriteJSON fails, we should not return an error

# Checklist:

### DIFF
--- a/api/utils/http.go
+++ b/api/utils/http.go
@@ -102,7 +102,11 @@ func ParseJSON(r io.Reader, v interface{}) error {
 // WriteJSON response an object in JSON encoding.
 func WriteJSON(w http.ResponseWriter, obj interface{}) error {
 	w.Header().Set("Content-Type", JSONContentType)
-	return json.NewEncoder(w).Encode(obj)
+	err := json.NewEncoder(w).Encode(obj)
+	if err != nil {
+		logger.Error("failed to write JSON response", "err", err)
+	}
+	return nil
 }
 
 // HandleGone is a handler for deprecated endpoints that returns HTTP 410 Gone.


api/utils/http.go
```diff
@@ -102,7 +102,11 @@ func ParseJSON(r io.Reader, v interface{}) error {
 // WriteJSON response an object in JSON encoding.
 func WriteJSON(w http.ResponseWriter, obj interface{}) error {
 	w.Header().Set("Content-Type", JSONContentType)
-	return json.NewEncoder(w).Encode(obj)
+	err := json.NewEncoder(w).Encode(obj)
+	if err != nil {
+		logger.Error("failed to write JSON response", "err", err)
```
darrenvechain, line 107: Not sure if we want to log this error. It seems to happen a lot in prod and I'm not convinced we actually care about it.

When I was reproducing this error locally it was mostly due to the client shutting down

darrenvechain: Basically what happens is:

- Client makes a request
- (I think) The client shuts down before it can receive a response
- WriteJSON writes the response code and body but it fails
- Error is thrown up to `WrapHandlerFunc` which tries to write a 500 - producing a double response header write

otherview: Makes sense to me, I wonder would it be worth logging a warning or debug or something instead of an error ?

otherview (approved):

api/utils/http.go
darrenvechain: I think warn will still produce unwanted logs. Will go with debug

otherview (approved):

api/utils/http.go
```diff
@@ -102,7 +102,11 @@ func ParseJSON(r io.Reader, v interface{}) error {
 // WriteJSON response an object in JSON encoding.
 func WriteJSON(w http.ResponseWriter, obj interface{}) error {
 	w.Header().Set("Content-Type", JSONContentType)
-	return json.NewEncoder(w).Encode(obj)
+	err := json.NewEncoder(w).Encode(obj)
```
freemanzMrojo, line 105: So now we always return nil if this function fails?

freemanzMrojo: Okay, just read the conversation above. Thanks

libotony (approved):

PR #1015, opened 2025-03-19
chore: replaced `go-bindata` by `embed`

freemanzMrojo:
# Description

Seems like we are using a pretty old version of `go-bindata` that generates deprecated code.

Removed the dependency with `go-bindata` to use `embed` instead. The files to embed are around 176KB so looks like it makes sense to have it done without compression.


.github/workflows/lint-go.yaml
```diff
@@ -18,12 +18,6 @@ jobs:
           go-version: '1.23'
           cache: false
```
darrenvechain: @freemanzMrojo is there a reason to remove this? I think its still important. Maybe we check the `compiled` dir now?

also, I think the compiled dir is in .gitignore

freemanzMrojo: I was trying things with the makefile to avoid having the compiled files now and I forgot to convert the PR into a draft

now it is still a draft but I tried what I wanted and I realised that it is just more convenient to have the files too (because docker mainly) so the GHA check is back.

It should be ready for review in a few seconds, sorry for the confusion

libotony:
@freemanzMrojo What are your thoughts on targeting this PR for `release/galactica`?

freemanzMrojo:
> @freemanzMrojo What are your thoughts on targeting this PR for `release/galactica`?

I do not mind, right now it is targeting `master` because regardless the forks `go-bindata` should be replaced.

libotony:
> > @freemanzMrojo What are your thoughts on targeting this PR for `release/galactica`?
> 
> I do not mind, right now it is targeting `master` because regardless the forks `go-bindata` should be replaced.

Yeah, just thinking our major work would be galactica, and we also have changes of builtin contracts in galactica. If you don't mind, let's bring that to galactica, thanks!

freemanzMrojo:
> > > @freemanzMrojo What are your thoughts on targeting this PR for `release/galactica`?
> > 
> > 
> > I do not mind, right now it is targeting `master` because regardless the forks `go-bindata` should be replaced.
> 
> Yeah, just thinking our major work would be galactica, and we also have changes of builtin contracts in galactica. If you don't mind, let's bring that to galactica, thanks!

Targeting now `release/galactica` @libotony

PR #1017, opened 2025-03-20
Fuzz tests

paologalligit:
# Description

This PR introduces fuzz testing for transaction, block and header encoding/decoding.

[GHIssue](https://github.com/vechain/protocol-board-repo/issues/464)

# Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,6 +60,11 @@ clean: #@ Clean the test and build cache and remove binaries
 test:| go_version_check #@ Run the tests
 	@go test -cover $(PACKAGES)
 
+fuzz:| go_version_check #@ Run the fuzz tests
+	@go test -fuzz=FuzzTransactionMarshalling -fuzztime=1m $(CURDIR)/tx
+	@go test -fuzz=FuzzBlockEncoding -fuzztime=1m $(CURDIR)/block
+	@go test -fuzz=FuzzHeaderEncoding -fuzztime=1m $(CURDIR)/block
+
 test-coverage:| go_version_check #@ Run the tests with coverage
 	@go test -race -coverprofile=coverage.out -covermode=atomic $(PACKAGES)
 	@go tool cover -html=coverage.out
--- a/block/block_test.go
+++ b/block/block_test.go
@@ -3,16 +3,17 @@
 // Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
 // file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
 
-package block_test
+package block
 
 import (
+	"fmt"
+	"math/big"
 	"testing"
 	"time"
 
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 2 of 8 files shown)


block/block_test.go
```diff
@@ -3,16 +3,17 @@
 // Distributed under the GNU Lesser General Public License v3.0 software license, see the accompanying
 // file LICENSE or <https://www.gnu.org/licenses/lgpl-3.0.html>
 
-package block_test
+package block
```
freemanzMrojo, line 6: Suggestion: I think it might make sense to keep `block_test` so we also check the method visibility as of now.

tx/tx_dynamic_fee.go
```diff
@@ -86,14 +86,14 @@ func (t *dynamicFeeTransaction) gasPriceCoef() uint8 {
 
 func (t *dynamicFeeTransaction) maxFeePerGas() *big.Int {
 	if t.MaxFeePerGas == nil {
-		return nil
+		return common.Big0
```
freemanzMrojo, line 89: I would check places where we are checking against `nil` (I have seen some for `maxPriorityFeePerGas()`.

For example:

https://github.com/vechain/thor/blob/release/galactica/runtime/resolved_tx.go#L51
https://github.com/vechain/thor/blob/release/galactica/txpool/validation.go#L37

Makefile
```diff
@@ -60,6 +60,11 @@ clean: #@ Clean the test and build cache and remove binaries
 test:| go_version_check #@ Run the tests
 	@go test -cover $(PACKAGES)
 
+fuzz:| go_version_check #@ Run the fuzz tests
```
freemanzMrojo, line 63: Question: How long do they take? Should we consider them for any pipeline?

freemanzMrojo (changes requested):

Makefile
paologalligit: If no argument is passed, the test will run forever or until an error is found. While the test run the fuzz engine will keep generating new inputs. At the moment I've set 1' timeout but we can change it. I think it makes sense to have a very long test run (previous PR was considering 30' each test) and not run these kind of tests in the pipeline every PR/merge action.

block/block_test.go
paologalligit: the go convention is to have the implementation and unit tests in the same package. If we want to have a set of tests interacting with the implementation via the public methods/structs exposed, they should be placed in a separated package named `block_test`

tx/tx_dynamic_fee.go
paologalligit: Yes, they're probably useless now. I'll remove them. Thanks

block/block_test.go
freemanzMrojo: Yep, just thought that it might make sense given the current status. Thanks.

Makefile
freemanzMrojo: Perfect, thanks for the explanation

freemanzMrojo (approved):

PR #1018, opened 2025-03-21
Increase limit per account

vanja-vechain:
# Description

Increases limit per account to 128 tx's

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:

### DIFF
--- a/api/node/node_test.go
+++ b/api/node/node_test.go
@@ -39,7 +39,7 @@ func initCommServer(t *testing.T) {
 		thorChain.Repo(),
 		txpool.New(thorChain.Repo(), thorChain.Stater(), txpool.Options{
 			Limit:           10000,
-			LimitPerAccount: 16,
+			LimitPerAccount: 128,
 			MaxLifetime:     10 * time.Minute,
 		}, &thor.NoFork),
 	)
--- a/cmd/thor/flags.go
+++ b/cmd/thor/flags.go
@@ -173,7 +173,7 @@ var (
 	}
 	txPoolLimitPerAccountFlag = cli.Uint64Flag{
 		Name:  "txpool-limit-per-account",
-		Value: 16,
+		Value: 128,
 		Usage: "set tx limit per account in pool",
 	}
 
--- a/cmd/thor/main.go
+++ b/cmd/thor/main.go
@@ -47,7 +47,7 @@ var (
 	copyrightYear        string
 	defaultTxPoolOptions = txpool.Options{
 		Limit:           10000,
-		LimitPerAccount: 16,
+		LimitPerAccount: 128,
 		MaxLifetime:     20 * time.Minute,
 	}
 )


paologalligit (approved):

YeahNotSewerSide (approved):

freemanzMrojo (approved):
LGTM, some minor comments

Not sure how thoroughly we want to go through this but I can see more related occurrences at:

1. https://github.com/vechain/thor/blob/release/galactica/api/subscriptions/pending_tx_test.go#L36
2. https://github.com/vechain/thor/blob/release/galactica/api/subscriptions/subscriptions_test.go#L236
3. https://github.com/vechain/thor/blob/release/galactica/api/transactions/transactions_benchmark_test.go#L188
4. https://github.com/vechain/thor/blob/release/galactica/api/transactions/transactions_test.go#L373

**Suggestion, for a separate ticket, I think other values should be evaluated too:**
After looking a bit into it I think it would make sense, given its usage across packages, to have something like a value in a var that gets overridden by the flag value, and that var is what we use instead of just the literal. This way we avoid missing occurrences in case more changes happen.

Something like this in terms of usage https://github.com/vechain/thor/blob/release/galactica/cmd/thor/main.go#L232 but with `var TxPoolLimitPerAccount = 128` and then we just use `TxPoolLimitPerAccount`. If going this path though it is important to remember the bootstrap order in Thor (otherwise we can get default values in some places where the flag is set with something else).

vanja-vechain:
> LGTM, some minor comments
> 
> Not sure how thoroughly we want to go through this but I can see more related occurrences at:
> 
> 1. https://github.com/vechain/thor/blob/release/galactica/api/subscriptions/pending_tx_test.go#L36
> 2. https://github.com/vechain/thor/blob/release/galactica/api/subscriptions/subscriptions_test.go#L236
> 3. https://github.com/vechain/thor/blob/release/galactica/api/transactions/transactions_benchmark_test.go#L188
> 4. https://github.com/vechain/thor/blob/release/galactica/api/transactions/transactions_test.go#L373
> 
> **Suggestion, for a separate ticket, I think other values should be evaluated too:** After looking a bit into it I think it would make sense, given its usage across packages, to have something like a value in a var that gets overridden by the flag value, and that var is what we use instead of just the literal. This way we avoid missing occurrences in case more changes happen.
> 
> Something like this in terms of usage https://github.com/vechain/thor/blob/release/galactica/cmd/thor/main.go#L232 but with `var TxPoolLimitPerAccount = 128` and then we just use `TxPoolLimitPerAccount`. If going this path though it is important to remember the bootstrap order in Thor (otherwise we can get default values in some places where the flag is set with something else).

Cheers @freemanzMrojo, yes was checking those and decided no to update since the bare tx limit was below 128 so it wouldn't have effect (of course except for the last one). There is an e2e test in the e2e repo to ensure 128 limit is respected.

PR #1019, opened 2025-03-24
Rename baseGas to baseFeePerGas as per VIP

paologalligit:
# Description

This PR renames the base fee value to `baseFeePerGas` when exposed via api

[GHIssue](https://github.com/vechain/protocol-board-repo/issues/489)

# Checklist:

### DIFF
--- a/api/fees/fees.go
+++ b/api/fees/fees.go
@@ -112,7 +112,7 @@ func (f *Fees) handleGetFeesHistory(w http.ResponseWriter, req *http.Request) er
 
 	return utils.WriteJSON(w, &FeesHistory{
 		OldestBlock:   oldestBlockRevision,
-		BaseFees:      baseFees,
+		BaseFeePerGas: baseFees,
 		GasUsedRatios: gasUsedRatios,
 	})
 }
--- a/api/fees/fees.go
+++ b/api/fees/fees.go
@@ -113,7 +113,7 @@ func (f *Fees) handleGetFeesHistory(w http.ResponseWriter, req *http.Request) er
 	return utils.WriteJSON(w, &FeesHistory{
 		OldestBlock:   oldestBlockRevision,
 		BaseFeePerGas: baseFees,
-		GasUsedRatios: gasUsedRatios,
+		GasUsedRatio:  gasUsedRatios,
 	})
 }
 
--- a/api/transactions/types.go
+++ b/api/transactions/types.go
@@ -82,7 +82,7 @@ type ReceiptMeta struct {
 
 // Receipt for json marshal
 type Receipt struct {
-	Type     uint8                 `json:"type,omitempty"`
+	Type     math.HexOrDecimal64   `json:"type,omitempty"`
 	GasUsed  uint64                `json:"gasUsed"`
 	GasPayer thor.Address          `json:"gasPayer"`
 	Paid     *math.HexOrDecimal256 `json:"paid"`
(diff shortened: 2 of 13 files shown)


paologalligit:
> Hey @paologalligit , as far as I see this PR addresses the attributes within the APIs but not within `Extension` for instance.
> 
> So is this enough?

Yes, the PR is renaming just the field exposed to the public, not internally. Have you seen any occurrences of `baseFee` in extension that is exposed?

freemanzMrojo:
> > Hey @paologalligit , as far as I see this PR addresses the attributes within the APIs but not within `Extension` for instance.
> > So is this enough?
> 
> Yes, the PR is renaming just the field exposed to the public, not internally. Have you seen any occurrences of `baseFee` in extension that is exposed?

Nope, just wanted to confirm my assumption 👍  Thanks.

I think we could modify the title of the PR too replacing `baseGasFeePerGas` by `baseFeePerGas`.

freemanzMrojo (changes requested):
Hey @paologalligit , after reviewing it again I think the `fees/history` endpoint should be adapted too so instead of `baseFees` we use `baseFeesPerGas` (in Eth it is singular but I find it weird, it is an array - feel free to use the singular version if you prefer so -)

libotony (approved):

api/transactions/types.go
```diff
@@ -82,7 +82,7 @@ type ReceiptMeta struct {
 
 // Receipt for json marshal
 type Receipt struct {
-	Type     uint8                 `json:"type,omitempty"`
+	Type     math.HexOrDecimal64   `json:"type,omitempty"`
```
darrenvechain, line 85: Think we still have this in the blocks package, for embedded txs, might be good to switch that too for consistency

```
Type:       trx.Type(),
```

darrenvechain: https://github.com/vechain/thor/blob/fc5b0405282ad858cb4bc2911a43c16fde692a49/api/blocks/types.go#L180

libotony: Uint8 is probably fine as other numbers are not returning in hex formats?

paologalligit: > Uint8 is probably fine as other numbers are not returning in hex formats?
This Pr wants to align the txType type to the same used in geth, which is returning a hex value

api/fees/fees.go
```diff
@@ -112,7 +112,7 @@ func (f *Fees) handleGetFeesHistory(w http.ResponseWriter, req *http.Request) er
 
 	return utils.WriteJSON(w, &FeesHistory{
 		OldestBlock:   oldestBlockRevision,
-		BaseFees:      baseFees,
+		BaseFeePerGas: baseFees,
```
freemanzMrojo, line 115: If using singular for this one the same criteria should apply to `GasUsedRatios` below (it should be just `GasUsedRatio`). See `Returns` [here](https://docs.metamask.io/wallet/reference/json-rpc-methods/eth_feehistory/) as a reference.

api/transactions/types.go
libotony: > > Uint8 is probably fine as other numbers are not returning in hex formats?
> 
> This Pr wants to align the txType type to the same used in geth, which is returning a hex value

Got what you mean, but after checking the existing API specs, seems we are not doing the same stuff like they're doing in geth, they are returning every number in hex string format. So I would suggest to align with our API spect, which is return number.

paologalligit: so, just to be on the same page, the new tx fields will be:
 - `txType`: int
 - `maxFeePerGas`: hex string
 - `maxPriorityFeePerGas`: hex string
correct?

libotony: Yeah, `maxFeePerGas` and `maxPriorityFeePerGas` are `big.Int`, that's why it is hex string.

api/fees/fees.go
otherview: Makes sense, applied that change 👍

freemanzMrojo (approved):

PR #1020, opened 2025-03-25
update encoding and signing hash func for tx

libotony:
# Description

+ Move typed tx encode/decode function to txbody
+ **Update signing hash function**
+ Return default value for incompatible functions

## Type of change

- [x] **Breaking change: tx signing hash changed to include tx type byte**

# Checklist:


tx/tx_legacy.go
```diff
@@ -145,3 +86,15 @@ func (t *legacyTransaction) evaluateWork(origin thor.Address) func(nonce uint64)
 		return r.Div(math.MaxBig256, r)
 	}
 }
+
+// Below are the methods that are not compatible with legacy transaction
```
otherview, line 90:
```suggestion
// Below are the methods that are not compatible with legacy transaction
// They return 0 as they are not meant to be used anywhere else
```

tx/tx_legacy.go
```diff
@@ -145,3 +86,15 @@ func (t *legacyTransaction) evaluateWork(origin thor.Address) func(nonce uint64)
...
+func (t *legacyTransaction) maxFeePerGas() *big.Int         { return common.Big0 }
+func (t *legacyTransaction) maxPriorityFeePerGas() *big.Int { return common.Big0 }
+
+func (t *legacyTransaction) encode(*bytes.Buffer) error {
+	panic("encode called on LegacyTx")
```
otherview, line 95: Should an error be returned instead ?

tx/tx_legacy.go
```diff
@@ -145,3 +86,15 @@ func (t *legacyTransaction) evaluateWork(origin thor.Address) func(nonce uint64)
...
+}
+
+func (t *legacyTransaction) decode([]byte) error {
+	panic("decode called on LegacyTx")
```
otherview, line 99: Should an error be returned instead ?

tx/tx_legacy.go
libotony: No necessarily, panic is a sanity check in development stage, logic has already covered in transaction.go that this won't be called

tx/tx_legacy.go
libotony: Added!

paologalligit (approved):
lgtm

tx/tx_legacy.go
otherview: Unless someone changes something in the transaction.go 🙃  
But I get the idea. I was coming from the point of, if the method was returning an error, why not leverage it.

libotony:
E2E Test https://github.com/vechain/thor-e2e-tests/pull/85

darrenvechain (approved):

PR #1028, opened 2025-03-28
feat: Rewards in `fee/history`

freemanzMrojo:
# Description

This PR adds support to:

- `rewardsPercentile` (request): List of percentiles (for instance, `20,50,90`) to organise the rewards to request.
- `reward` (response): Array of arrays with the rewards by percentile (in the example above, it will be an array of 3-element arrays)

It is based on this.

Some things to consider:

1. We are limiting to 100 percentiles at most, same as go-ethereum (so `rewardPercentiles` could have 100 elements)
2. We split cache handling from reward calculation. This is due to a request from the user with different sizes for `rewardPercentiles` (I can request `20,30` for block 30 and `50` later on for the same block, if using the cache directly for this we will be storing the first call so 2 values for this block).

Closes https://github.com/vechain/protocol-board-repo/issues/508

### DIFF
--- a/api/fees/data.go
+++ b/api/fees/data.go
@@ -20,11 +20,10 @@ import (
 	"github.com/vechain/thor/v2/thor"
 )
 
-type FeeCacheEntry struct {
-	parentBlockID thor.Bytes32
-	baseFee       *hexutil.Big
-	gasUsedRatio  float64
-	cachedRewards *rewards
+type FeesData struct {
+	repo   *chain.Repository
+	cache  *cache.PrioCache
+	stater *state.Stater
 }
 
 type rewardItem struct {
@@ -37,10 +36,11 @@ type rewards struct {
 	totalGasUsed uint64
 }
 
-type FeesData struct {
-	repo   *chain.Repository
-	cache  *cache.PrioCache
-	stater *state.Stater
+type FeeCacheEntry struct {
+	parentBlockID thor.Bytes32
+	baseFee       *hexutil.Big
+	gasUsedRatio  float64
+	cachedRewards *rewards
 }
 
 func newFeesData(repo *chain.Repository, stater *state.Stater, fixedCacheSize int) *FeesData {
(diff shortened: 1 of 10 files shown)


api/fees/fees.go
```diff
@@ -42,70 +45,111 @@ func calcPriorityFee(baseFee *big.Int, priorityPercentile int64) *big.Int {
...
 	if blockCount == 0 {
-		return 0, nil, utils.BadRequest(errors.New("invalid blockCount, it should not be 0"))
+		return nil, utils.BadRequest(errors.New("invalid blockCount, it should not be 0"))
 	}
 
-	// Validate newestBlock
+	return &blockCount, nil
+}
+
+func (f *Fees) validateNewestBlock(req *http.Request, blockCount *uint64) (*chain.BlockSummary, error) {
```
paologalligit, line 88: If the only reason why the block count is passed by its pointer is we can change it, can we return it instead of changing it silently?

api/fees/data.go
```diff
@@ -83,14 +114,98 @@ func (fd *FeesData) getOrLoadFees(blockID thor.Bytes32) (*FeeCacheEntry, error)
...
+		return a.reward.Cmp(b.reward)
+	})
+
+	return &rewards{
+		items:        items,
+		totalGasUsed: header.GasUsed(),
+	}, nil
+}
+
+func (fd *FeesData) calculateRewards(cachedRewards *rewards, rewardPercentiles []float64) []*hexutil.Big {
```
paologalligit, line 183: Do you think it's worth to add some more unit tests for this function alone? Maybe with some edge cases like cachedRewards is nil, rewardPercentiles is [] and other examples

freemanzMrojo: Sure thing @paologalligit , there are currently integration tests that should cover indirectly both scenarios (essentially all tests at `fees_test.go` except for `TestRewardPercentiles` do not pass `rewardPercentiles`) but there is no harm in having unit tests also here, will do it.

Thanks.

api/fees/fees.go
freemanzMrojo: Yeah, that was the reason. Modified and pushed, thanks.

api/fees/data.go
freemanzMrojo: Done now

paologalligit (approved):

PR #1038, opened 2025-04-08
feat(legacy tx): error message when base fee is greater than baseGasPrice

freemanzMrojo:
# Description

Title is self-explanatory. Also reviewed the checks at `txpool/validation.go` but I think the errors are good there since both tx fee values are 0 in those cases for legacy tx.

Closes https://github.com/vechain/protocol-board-repo/issues/537

### DIFF
--- a/consensus/fork/galactica.go
+++ b/consensus/fork/galactica.go
@@ -20,9 +20,9 @@ import (
 var (
 	// errBaseFeeTooHighForLegacyTx is returned if the base fee is too high for a legacy tx
 	errBaseFeeTooHighForLegacyTx = errors.New("base fee too high for legacy tx, use dynamic fee tx or retry later")
-	// errMaxFeePerGasTooLow is returned if the transaction max fee is less than
+	// errBaseFeeTooHighForDynamicFeeTx is returned if the transaction max fee is less than
 	// the base fee of the block.
-	errMaxFeePerGasTooLow = errors.New("max fee per gas is less than block base fee")
+	errBaseFeeTooHighForDynamicFeeTx = errors.New("max fee per gas is less than block base fee")
 	// ErrBaseFeeNotSet is returned if the base fee is not set after the Galactica fork
 	ErrBaseFeeNotSet = errors.New("base fee not set after galactica")
 )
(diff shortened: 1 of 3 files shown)


YeahNotSewerSide (approved):
lgtm

consensus/fork/galactica.go
```diff
@@ -18,11 +18,13 @@ import (
 )
 
 var (
+	// errBaseFeeTooHighForLegacyTx is returned if the base fee is too high for a legacy tx
+	errBaseFeeTooHighForLegacyTx = errors.New("base fee too high for legacy tx, use dynamic fee tx or retry later")
```
paologalligit, line 22: do you think we can have a similar error naming? Since they refer to the same root cause. Like `errBaseFeeTooHighForLegacyTx` and `errBaseFeeTooHighForDynamicFeeTx`

consensus/fork/galactica_test.go
```diff
@@ -455,7 +455,7 @@ func TestValidateGalacticaTxFee(t *testing.T) {
 			tx:           tx.NewBuilder(tx.TypeLegacy).GasPriceCoef(255).Build(),
 			baseFee:      defaultBaseFee,
 			baseGasPrice: new(big.Int).Sub(defaultBaseFee, common.Big1),
-			wantErr:      errors.New("max fee per gas is less than block base fee"),
+			wantErr:      errors.New("base fee too high for legacy tx, use dynamic fee tx or retry later"),
```
paologalligit, line 458: Here we could replace the `wantErr: errors.New(...` with `wantErr: errBaseFeeTooHighForLegacyTx` directly. Also on line 465 with the dyn error msg

consensus/validator_test.go
```diff
@@ -118,7 +119,30 @@ func TestValidateBlock(t *testing.T) {
 				s, r, err := c.verifyBlock(blk, state, 0)
 				assert.Nil(t, s)
 				assert.Nil(t, r)
-				assert.Equal(t, fork.ErrMaxFeePerGasTooLow, err)
+				expectedErr := errors.New("max fee per gas is less than block base fee")
```
paologalligit, line 122: Also here, is there any reason why we want the error msg in the fork package to be private and not public anymore? We could expect the error from the fork package here and on line 145

consensus/fork/galactica.go
freemanzMrojo: Done, thanks

consensus/fork/galactica_test.go
freemanzMrojo: Done, thanks!

consensus/validator_test.go
freemanzMrojo: As far as I see it is only used by its file and by this tests.

IMO the `var` should be package-scoped since there is no actual feature outside of the `fork` package using it.

If you feel strongly about keeping it public I do not mind changing it back, however the new error will follow the same criteria (so public too).

paologalligit (approved):

PR #1039, opened 2025-04-08
refactor: legacy base gas price retrieval within `ValidateGalacticaTxFee`

freemanzMrojo:
# Description

Since we always use the state value for `baseGasPrice` to validate the fees, this PR aims for atomising the functionality within the `ValidateGalacticaTxFee` function.


otherview (commented):
I think refactoring of the`flow.isEffectivePriorityFeeTooLow` is good because the flow already has the state, no point passing it a value that's derived from it again.

With regard to `ValidateGalacticaTxFee` what's the concern of taking the state vs the value it uses ?

freemanzMrojo:
> I think refactoring of the`flow.isEffectivePriorityFeeTooLow` is good because the flow already has the state, no point passing it a value that's derived from it again.
> 
> With regard to `ValidateGalacticaTxFee` what's the concern of taking the state vs the value it uses ?

Hi @otherview , no concerns. I just noticed that for every case (except for the test), when we call `ValidateGalacticaTxFee` we also call the function to return the `baseGasPrice` with the purpose of passing that value (we always need it within that function).

The goal of this PR is to put them together. This is a case where IMO the code is clearer:

- Nowadays

```
		if header.BaseFee() != nil {
			baseGasPrice, err := builtin.Params.Native(state).Get(thor.KeyBaseGasPrice)
			if err != nil {
				return nil, nil, err
			}
			if err := fork.ValidateGalacticaTxFee(tx, header.BaseFee(), baseGasPrice); err != nil {
				return nil, nil, err
			}
		}
```

- With this PR

```
		if header.BaseFee() != nil {
			if err := fork.ValidateGalacticaTxFee(tx, state, header.BaseFee()); err != nil {
				return nil, nil, err
			}
		}
```

otherview:
Yeah, I see what you mean, it's wrapping up the base fee retrieval logic. It's starting to look as it might be pushing the logic downstream. Now the state is something the fork package must understand. Which now means the fork package must handle state to run the tests.

My preference would be to maintain the function as simple as possible with simple types. Not sure that it warrants a refactor. But no strong feelings here, happy to get another opinion and go with the popular take on it.

I guess one possibility could be to have something like:

```
func ValidateGalacticaTxFee(tr *tx.Transaction, state *state.State, baseFee *big.Int) error {
	baseGasPrice, err := builtin.Params.Native(state).Get(thor.KeyBaseGasPrice)
	if err != nil {
		return err
	}
	return validateGalacticaTxFee(tr, state, baseFee, baseGasPrice)
	}

func validateGalacticaTxFee(tr *tx.Transaction,baseGasPrice, baseFee *big.Int) error {
      blablabla
```

And then have the tests run against the private method?

freemanzMrojo:
> Yeah, I see what you mean, it's wrapping up the base fee retrieval logic. It's starting to look as it might be pushing the logic downstream. Now the state is something the fork package must understand. Which now means the fork package must handle state to run the tests.
> 
> My preference would be to maintain the function as simple as possible with simple types. Not sure that it warrants a refactor. But no strong feelings here, happy to get another opinion and go with the popular take on it.
> 
> I guess one possibility could be to have something like:
> 
> ```
> func ValidateGalacticaTxFee(tr *tx.Transaction, state *state.State, baseFee *big.Int) error {
> 	baseGasPrice, err := builtin.Params.Native(state).Get(thor.KeyBaseGasPrice)
> 	if err != nil {
> 		return err
> 	}
> 	return validateGalacticaTxFee(tr, state, baseFee, baseGasPrice)
> 	}
> 
> func validateGalacticaTxFee(tr *tx.Transaction,baseGasPrice, baseFee *big.Int) error {
>       blablabla
> ```
> 
> And then have the tests run against the private method?

Yeah, the "main" logic of the method can be extracted to a private method, will do that

freemanzMrojo:
@otherview I extracted the logic into the private method but the tests still run over the entire method since otherwise we wont be covering the state retrieval (or we would need specific tests just for that).

If it is a hard requirement to have also tests for the private method, even though it is covered now, just let me know and I will add them.

otherview:
This is more of a personal preference, I believe that in golang it's typically preferred to pass in simple values for methods, this allows for the testing and complexity of the function to be reduced. 

This implementation works, perhaps the most important part of testing is the private method. I'm fine with only testing the private method, as retrieving data from the state should be covered in other parts of the test suite.

freemanzMrojo:
> This is more of a personal preference, I believe that in golang it's typically preferred to pass in simple values for methods, this allows for the testing and complexity of the function to be reduced.
> 
> This implementation works, perhaps the most important part of testing is the private method. I'm fine with only testing the private method, as retrieving data from the state should be covered in other parts of the test suite.

Ok, done now that way @otherview , thanks.

otherview (approved):
lgtm 👍

PR #1054, opened 2025-04-23
improve code structure

libotony:
# Description

This PR improves code structure in several packages, details:

+ Introduced concept `EffectiveGasPrice` and `EffectivePriorityFeePerGas` in tx package
+ Moved some tx basic validation to `runtime.ReolveTransaction`
+ Renamed package `consensus/fork` to `consensus/upgrade/galactica`
+ Moved legacy tx only function to itself, marked some legacytx-only methods
+ Added a `params` package in `txpool` to help caching dynamic params like `baseFee`, `legacyTxBaseGasPrice`

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing E2E tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have not added any vulnerable dependencies to my code

PR #1065, opened 2025-05-05
Reduce the number of state calls when processing blocks

otherview:
# Description

This is a PR comes from https://github.com/vechain/thor/pull/1054
It reduces the number of state calls done to retrieve the `thor.KeyLegacyTxBaseGasPrice`. 
Instead of once per tx, it's now done once per flow/block.

Fixes # (issue)

## Type of change

* Docker Version:

# Checklist:

### DIFF
--- a/packer/packer.go
+++ b/packer/packer.go
@@ -127,7 +127,10 @@ func (p *Packer) Schedule(parent *chain.BlockSummary, nowTimestamp uint64) (flow
 		}
 	}
 
-	var baseFee *big.Int
+	var legacyTxBaseGasPrice, baseFee *big.Int
+	if legacyTxBaseGasPrice, err = builtin.Params.Native(state).Get(thor.KeyLegacyTxBaseGasPrice); err != nil {
+		return nil, err
+	}
 
 	if parent.Header.Number()+1 >= p.forkConfig.GALACTICA {
 		baseFee = fork.CalcBaseFee(&p.forkConfig, parent.Header)
@@ -147,7 +150,7 @@ func (p *Packer) Schedule(parent *chain.BlockSummary, nowTimestamp uint64) (flow
 		},
 		p.forkConfig)
 
-	return newFlow(p, parent.Header, rt, features), nil
+	return newFlow(p, parent.Header, rt, features, legacyTxBaseGasPrice), nil
 }
 
 // Mock create a packing flow upon given parent, but with a designated timestamp.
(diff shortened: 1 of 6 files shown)


packer/packer.go
```diff
@@ -166,7 +169,11 @@ func (p *Packer) Mock(parent *chain.BlockSummary, targetTime uint64, gasLimit ui
 		gl = p.gasLimit(parent.Header.GasLimit())
 	}
 
-	var baseFee *big.Int
+	var err error
+	var legacyTxBaseGasPrice, baseFee *big.Int
+	if legacyTxBaseGasPrice, err = builtin.Params.Native(state).Get(thor.KeyLegacyTxBaseGasPrice); err != nil {
```
darrenvechain, line 174: Will this break if value of `KeyLegacyTxBaseGasPrice` changes?

otherview: It shouldn't - from my understanding, the `KeyLegacyTxBaseGasPrice` changes should only be applicable in the next block.

freemanzMrojo (approved):

packer/packer.go
libotony: I am good with both approaches, but previous code defined the logic that `KeyLegacyTxBaseGasPrice` takes effect immediately right after the changing tx is executed. So this will break the existing chain if syncing from start with new code. 
And this improvement is minor since we have cache layer in the stateDB, multiple access does not result multiple IO. 
BTW, I have gone the same way as you did @otherview but turns out not working.

otherview: > previous code defined the logic that KeyLegacyTxBaseGasPrice takes effect immediately right after the changing tx is executed. So this will break the existing chain if syncing from start with new code.

I see. So in the tx, the `KeyLegacyTxBaseGasPrice` and the balance deduced from other accounts on the same block (txs that are listed after the change-base-gas tx). And this would be a different state root from the case where the `KeyLegacyTxBaseGasPrice` only takes effect on the next block.

otherview: Yeah, I think we're ok to remove an extra state check in the `Adopt().fork.ValidateGalacticaTxFee` but that's about it.

Added a comment explaining the rationale behind it.
The tldr is that the `Adopt().fork.ValidateGalacticaTxFee` checks for if the tx has enough gas price to pay for the block base fee. It's only at the `Adopt().f.runtime.ExecuteTransaction(t)` that any state changes are applied, and in that method the state is refreshed to account for any changes that tx's might have produced.

otherview: Was able to create a scenario where a tx when it shoudlnt.

```
// Current block DynFee Base Gas Price = KeyLegacyTxBaseGasPrice
// tx1 = LegacyTx doing something with BaseGasCoef = 0 				OK! State = same and f.legacyTxBaseGasPrice = same
// tx2 = Otto changing KeyLegacyTxBaseGasPrice from 1 to 0.5		OK! State = Same and f.legacyTxBaseGasPrice = same
// tx3 = LegacyTx doing something with BaseGasCoef = 0 				?   State = DIFFERENT and f.legacyTxBaseGasPrice = same
```

PR #1095, opened 2025-06-02
Refactor bind

otherview:
# Description

This PR introduces a more intuitive API for contract interactions, building upon the existing foundation to provide a clearer decoupling of concerns and better experience.

The new API provides distinct interfaces for different types of operations:
- `MethodBuilder` as the entry point for all contract method interactions
- `CallBuilder` for read operations (view/pure functions)
- `SendBuilder` for write operations (state-changing functions)
- `FilterBuilder` for event filtering
- `Signer` interface for transaction signing

This separation prevents semantically incorrect operations that were possible previously, such as:
```go
// Old API - could lead to confusion
contract.Call("transfer", to, amount).Send() // Mixing read and write operations

// New API - clear intent
contract.Method("transfer", to, amount).Send().WithSigner(signer).Send() // Explicit write operation
```

The new `MethodBuilder` provides a nicer starting point for all contract interactions, making it easier to choose the appropriate operation type:

```go
// Create an operation once, use it multiple ways
op := contract.MethodBuilder("transfer", to, amount)

// Option 1: Read/Simulate the operation
result, err := op.Call().Execute()           // Simple call
err = op.Call().Into(&balance)              // Call and unpack result
simResult, err := op.Call().Simulate(caller) // Simulate with specific caller

// Option 2: Execute the operation
tx, err := op.Send().
    WithSigner(signer).
    WithOptions(&TxOptions{Gas: &gas}).
    Send()                                  // Send without waiting

receipt, tx, err := op.Send().
    WithSigner(signer).
    Receipt(ctx)                            // Send and wait for receipt
```

Fixes # [(issue)](https://github.com/vechain/otherviews-workboard/issues/181)

# Checklist:


thorclient/bind/contract.go
```diff
@@ -0,0 +1,93 @@
...
+
+// contract is the concrete implementation of Contract.
+type contract struct {
+	client *thorclient.Client
+	abi    *abi.ABI
+	addr   *thor.Address
+}
+
+// NewContract creates a new contract instance with the given client, ABI data and address.
+func NewContract(client *thorclient.Client, abiData []byte, address *thor.Address) (Contract, error) {
```
darrenkel, line 45: nit: using parent package can get quite messy. IMO we should either use httpclient or move this package so its not a child

thorclient/bind/contract.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/vechain/thor/v2/thor"
+	"github.com/vechain/thor/v2/thorclient"
+)
+
+// Contract is the main interface for contract interactions.
+// It provides a unified entry point for all contract operations.
+type Contract interface {
+	// Method creates a new method builder for the specified method and arguments.
+	Method(method string, args ...any) MethodBuilder
```
libotony, line 22:
```suggestion
	Method(method string) MethodBuilder
```
Arguments can be passed for building the call/send builder
```
method := contract.MethodBuilder("transfer", to, amount)

method.Call(to, amount).Execute()
method.Send(to, amount).Send()
```

thorclient/bind/contract.go
```diff
@@ -0,0 +1,93 @@
...
+	return &contract{
+		client: client,
+		abi:    &contractABI,
+		addr:   address,
+	}, nil
+}
+
+// Method implements Contract.Method.
+func (c *contract) Method(method string, args ...any) MethodBuilder {
+	return &methodBuilder{
```
libotony, line 62: Should verify method name exists here?

thorclient/bind/contract.go
```diff
@@ -0,0 +1,93 @@
...
+		contract: c,
+		method:   method,
+		args:     args,
+		vet:      big.NewInt(0),
+	}
+}
+
+// FilterEvent implements Contract.Event.
+func (c *contract) FilterEvent(eventName string) FilterBuilder {
+	return &filterBuilder{
```
libotony, line 72: Should verify event name exists here?

thorclient/bind/call.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Into unpacks the result into the provided interface.
+	Into(result any) error
+
+	// Execute performs the call and returns the raw result.
+	Execute() (*accounts.CallResult, error)
+
+	// Simulate performs a call simulation with the specified caller.
+	Simulate(caller *thor.Address) (*accounts.CallResult, error)
+}
```
libotony:
```suggestion
type CallBuilder interface {
	// AtRevision sets the revision for the call.
	AtRevision(rev string) CallBuilder
	
	Caller(caller *thor.Address) CallBuilder

	// Into unpacks the result into the provided interface.
	ExecuteInto(result any) error

	// Execute performs the call and returns the raw result.
	Execute() (*accounts.CallResult, error)
}
```

How about this？

libotony: For `ExecuteInto`, how about `ExecuteInto[T any]()(T, error)` ? Then no need to worry if the passing one is a pointer?

thorclient/bind/contract.go
darrenkel: IMO its a better strategy to do this later, during the execution. Because otherwise you have lots of `res, err := (...)`, just leads to lots of duplication for not much benefit

thorclient/bind/send.go
```diff
@@ -0,0 +1,198 @@
...
+	if _, err = b.op.contract.client.SendTransaction(transaction); err != nil {
+		return nil, err
+	}
+
+	return transaction, nil
+}
+
+// Receipt implements SendBuilder.Receipt.
+func (b *sendBuilder) Receipt(ctx context.Context) (*transactions.Receipt, *tx.Transaction, error) {
+	transaction, err := b.IssueTx()
```
darrenkel: Might be a bit confusing for clients if they call `IssueTx` and then `Receipt` ? ie, they end up sending the tx twice even though they only want 1 

This is the reason I cached the tx previously, if built/ sent

thorclient/bind/contract.go
otherview: Yep, it would be nice, but would break the method chaining.

thorclient/bind/contract.go
otherview: Same as above, keeps it with the same pattern where only at execution are the validations done.

thorclient/bind/call.go
otherview: I've update the names and used the Caller. I agree, it makes it better aligned with the current pattern. Thanks!

>For ExecuteInto, how about ExecuteInto[T any]()(T, error) ? Then no need to worry if the passing one is a pointer?

That doesn't seem to work as `Interface method must have no type parameters`

Would have to create a new typed builder such as:
`Method[T any](method string, args ...any) *TypedCallBuilder[T]`

This looks fine, but with the extra code it would be required, I would consider implementing it at a later stage if it becomes a problem.

thorclient/bind/send.go
otherview: Good point - changing it !

thorclient/bind/contract.go
otherview: Discussed offline 👍

runtime/runtime_test.go
```diff
@@ -772,7 +772,7 @@ func TestExecuteTransaction(t *testing.T) {
 		Build()
 	legacyTx = tx.MustSign(legacyTx, genesis.DevAccounts()[0].PrivateKey)
 
-	t.Run("Receipt check with legacy tx before galactica fork", func(t *testing.T) {
+	t.Run("SubmitAndConfirm check with legacy tx before galactica fork", func(t *testing.T) {
```
darrenkel: Bad Jetbrains refactor I think!

otherview: Yep! good catch 🙏

darrenvechain (approved):

thorclient/bind/contract.go
otherview: Yeah, I think you're right. It's possible this can lead to a circular dependency.
I think this raises concerns with thorbuilder packaging as it's a struct and should probably be an interface. 
Which would enable for a thorclient implementation based on testchain.

Given it's not creating an issue atm, and the solution my take a bit of time, I think it's ok to keep it until either a problem occours or theres a refactor of thorclient.

PR #1105, opened 2025-06-04
Improve logdb inmem

otherview:
# Description

- Optimized in-memory SQLite database configuration in `logdb/logdb.go` for better test performance
- Optimized SQLite settings in `logdb/logdb.go` for in-memory operations:
  - `synchronous=OFF`
  - `journal_mode=MEMORY`
  - `cache_size=-2000`
  - `temp_store=MEMORY`

Fixes # (issue)

## Type of change

# Checklist:

### DIFF
--- a/logdb/logdb.go
+++ b/logdb/logdb.go
@@ -70,8 +70,6 @@ func New(path string) (logDB *LogDB, err error) {
 		wconn:         wconn1,
 		wconnSyncOff:  wconn2,
 		stmtCache:     newStmtCache(db),
-		// todo investigate having a dedicated writer similar to the InMem
-		//writer:        &Writer{conn: wconn1, stmtCache: newStmtCache(db)},
 	}, nil
 }
 
--- a/logdb/logdb.go
+++ b/logdb/logdb.go
@@ -10,6 +10,7 @@ import (
 	"database/sql"
 	"fmt"
 	"math/big"
+	"time"
 
 	sqlite3 "github.com/mattn/go-sqlite3"
 	"github.com/vechain/thor/v2/block"
@@ -28,6 +29,7 @@ type LogDB struct {
 	wconn         *sql.Conn
 	wconnSyncOff  *sql.Conn
 	stmtCache     *stmtCache
+	writer        *Writer
 }
 
 // New create or open log db at given path.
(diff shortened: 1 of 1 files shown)


logdb/logdb.go
```diff
@@ -68,12 +70,81 @@ func New(path string) (logDB *LogDB, err error) {
 		wconn:         wconn1,
 		wconnSyncOff:  wconn2,
 		stmtCache:     newStmtCache(db),
+		// todo investigate having a dedicated writer similar to the InMem
+		//writer:        &Writer{conn: wconn1, stmtCache: newStmtCache(db)},
```
darrenkel: do we need to resolve this?

otherview: Good point - I meant to capture this in a ticket. 
Any changes to the logdb should have a performance test or something that measures the value of the change.
I dont think there's a need to make changes atm, but its a worthy investigation feature.

logdb/logdb.go
```diff
@@ -72,8 +74,75 @@ func New(path string) (logDB *LogDB, err error) {
...
 
 // NewMem create a log db in ram.
+// Unlike the file-based database, this implementation:
+// 1. Uses synchronous=off for faster writes in memory
+// 2. Uses a single connection with minimal transaction overhead
+// 3. Optimizes for in-memory performance
 func NewMem() (*LogDB, error) {
-	return New("file::memory:")
+	// Generate a unique database name to prevent concurrent access issues
+	dbName := fmt.Sprintf("file:memdb_%d?mode=memory&cache=shared&_txlock=immediate&synchronous=off&journal_mode=memory", time.Now().UnixNano())
```
libotony: nit; In an extreme concurrent scenario, nano timestamp can also be the same, how about random bytes in 4 or 6?

logdb/logdb.go
```diff
@@ -72,8 +74,75 @@ func New(path string) (logDB *LogDB, err error) {
...
+	// - synchronous=off: Fastest mode for in-memory database
+	// - journal_mode=memory: Use in-memory journal for better performance
+	db, err := sql.Open("sqlite3", dbName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set optimized PRAGMA settings for in-memory performance
+	if _, err := db.Exec(`
+		PRAGMA synchronous = OFF;
```
libotony: `synchronous` and `journal_mode` already declared in the connection string, duplicated here?

logdb/logdb.go
```diff
@@ -28,6 +29,7 @@ type LogDB struct {
 	wconn         *sql.Conn
 	wconnSyncOff  *sql.Conn
 	stmtCache     *stmtCache
+	writer        *Writer
```
libotony: Writer is not used?

logdb/logdb.go
```diff
@@ -72,8 +74,75 @@ func New(path string) (logDB *LogDB, err error) {
...
+	// - journal_mode=memory: Use in-memory journal for better performance
+	db, err := sql.Open("sqlite3", dbName)
+	if err != nil {
+		return nil, err
+	}
+
+	// Set optimized PRAGMA settings for in-memory performance
+	if _, err := db.Exec(`
+		PRAGMA synchronous = OFF;
+		PRAGMA journal_mode = MEMORY;
```
vanja-vechain: Do you see any benefits in adding optimize here 
PRAGMA optimize;
PRAGMA optimize(MASK);
PRAGMA schema.optimize;
PRAGMA schema.optimize(MASK)?

logdb/logdb.go
otherview: Good point - fixed 👍

logdb/logdb.go
otherview: yep ! Removed 👍

logdb/logdb.go
otherview: Thats a good question - Had a deeper look into the optimizations flags

```
-- Database optimizations:
-- 0x0001 (ANALYZE): Updates table statistics for better query planning
-- 0x0002 (REINDEX): Rebuilds all indexes to ensure optimal performance
-- 0x0004 (VACUUM): Not used as it's less relevant for in-memory DB
-- 0x0008 (INTEGRITY_CHECK): Not used to avoid performance overhead
-- 0x0010 (STATS): Not used as ANALYZE provides sufficient statistics
```

Added mask = 0x003 ( Analyze + Reindex ) given the others are not relevant and the bind + builtin tests passed.

Then I wondered, do we really need these optimizations for the NewMemDb ?
So I removed all optimizations and the tests still look ok.
So I removed the writer separated connections and the tests still look ok.

It looks like this PR went a bit too deep, the per-instance connection + standard optimizations seem to do the trick.
I'm removing the extra code as it's not reflecting the added value.

Many thanks for suggesting this 🙏

logdb/logdb.go
otherview: Removed 👍

libotony (approved):

PR #1127, opened 2025-06-26
use receipt to calculate effective priority gas price in fees history…

libotony:
… API

# Description

To decrease  computation per tx.

# Checklist:

### DIFF
--- a/api/fees/data.go
+++ b/api/fees/data.go
@@ -12,10 +12,8 @@ import (
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/vechain/thor/v2/block"
-	"github.com/vechain/thor/v2/builtin"
 	"github.com/vechain/thor/v2/cache"
 	"github.com/vechain/thor/v2/chain"
-	"github.com/vechain/thor/v2/consensus/upgrade/galactica"
 	"github.com/vechain/thor/v2/state"
 	"github.com/vechain/thor/v2/thor"
 )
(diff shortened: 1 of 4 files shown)


libotony:
Needs to wait #1126

txpool/tx_object.go
```diff
@@ -63,7 +65,7 @@ func (o *txObject) Payer() *thor.Address {
 	return o.payer
 }
 
-func (o *txObject) Executable(chain *chain.Chain, state *state.State, headBlock *block.Header, forkConfig *thor.ForkConfig) (bool, error) {
+func (o *txObject) Executable(chain *chain.Chain, state *state.State, headBlock *block.Header, forkConfig *thor.ForkConfig, baseFee *big.Int) (bool, error) {
```
otherview: Looks like most checks are done on the next block, but the base fee is referencing the current block.

txpool/tx_pool.go
```diff
@@ -267,7 +266,7 @@ func (p *TxPool) add(newTx *tx.Transaction, rejectNonExecutable bool, localSubmi
 		}
 
 		state := p.stater.NewState(headSummary.Root())
-		executable, err := txObj.Executable(p.repo.NewChain(headSummary.Header.ID()), state, headSummary.Header, p.forkConfig)
+		executable, err := txObj.Executable(p.repo.NewChain(headSummary.Header.ID()), state, headSummary.Header, p.forkConfig, p.baseFeeCache.Get(headSummary))
```
otherview: It's likely that `p.baseFeeCache.Get(headSummary)` might need to be reworked here to provide the `nextBlock` base gas price.

In `Executable` the calculations are done with block+1, but the base gas price is not touched.

txpool/tx_object.go
libotony: `CalcBaseFee` is calculating for the next block, you put the parent header then it calculates the next block's fee.

txpool/tx_pool.go
libotony: `BaseGasPrice` is dynamic to txs that changing it, not to block number/time, so for the 'next' block you use parent block's base gas price.

libotony: Hmmm, I was talking about `BaseGasPrice` but you were referring to `BaseFee`, check my reply on the previous comment, it covered the `BaseFee` stuff.

txpool/tx_object.go
otherview: Makes sense, thanks !

txpool/tx_pool.go
otherview: Yep yep, resolved above 👍

otherview (approved):

otherview (approved):

PR #1160, opened 2025-07-09
Vanja/feat/675 double signing evidence storage implementation

vanja-vechain:
# Description

Evidence storage implementation. Double signing evidence is stored in block header in following blocks

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:


cmd/thor/node/node.go
```diff
@@ -456,6 +484,41 @@ func (n *Node) processBlock(newBlock *block.Block, stats *blockStats) (bool, err
 	return *isTrunk, nil
 }
 
+func (n *Node) validateEvidence(evidence *[][]byte) error {
```
YeahNotSewerSide: Hey, I have some questions about this function, as I understand it relies on getting the blocks by IDs from the repository, but is it possible that block specified in evidence is not in node's local repo? Should the headers of double signed blocks themselves be stored as an evidence?

vanja-vechain: Intention here was to reduce the size of the header but I get your concern, let me try to re-work it

chain/repository.go
```diff
@@ -447,3 +449,28 @@ func (r *Repository) IsNotFound(err error) bool {
...
+	} else {
+		result, found := r.caches.doubleSig.Get(r.caches.doubleSig.Keys()[0])
+		if !found {
+			return nil
+		} else {
+			evidence := result.([][]byte)
+			return &evidence
+		}
+	}
+}
```
freemanzMrojo: Quick question, if we always get the first key, why do we store others? Are they used in a separate place of the code?

Also a suggestion to simplify these conditions:

```suggestion
func (r *Repository) GetDoubleSigEvidence() *[][]byte {
	if r.caches.doubleSig.Len() == 0 {
		return nil
	}

	result, found := r.caches.doubleSig.Get(r.caches.doubleSig.Keys()[0])
	if !found {
		return nil
	}

	evidence := result.([][]byte)
	return &evidence
}
```

cmd/thor/node/node.go
```diff
@@ -429,6 +448,15 @@ func (n *Node) processBlock(newBlock *block.Block, stats *blockStats) (bool, err
 		}
 		stats.UpdateProcessed(1, len(receipts), execElapsed, commitElapsed, realElapsed, newBlock.Header().GasUsed())
 
+		if isPos && evidence != nil && len(*evidence) > 1 {
```
freemanzMrojo, line 451: Suggestion: Maybe I am missing context because this method is pretty long, but can this block of code be included above (along with the new code where `evidence` is created) so everything is together?

Actually, once you get `conflictingBlocks` you are good to go to perform the 3 new blocks of code about evidence of double-signing right? Maybe they could be together and also extracted into a new method 👍🏼

packer/flow.go
```diff
@@ -248,6 +248,10 @@ func (f *Flow) Pack(privateKey *ecdsa.PrivateKey, newBlockConflicts uint32, shou
 		builder.COM()
 	}
 
+	if newBlockConflicts > 0 && f.PosActive {
```
freemanzMrojo: Suggestion: Maybe we can check `newBlockConflicts` within the `f.PosActive` condition above? So we can delete this one here and avoid checking twice against `f.PosActive`. The `builder` would need to be created before the first check though.

cmd/thor/node/packer_loop.go
```diff
@@ -204,6 +209,15 @@ func (n *Node) pack(flow *packer.Flow) (err error) {
 			logger.Trace("bandwidth updated", "gps", v)
 		}
 
+		if evidence != nil && flow.PosActive {
```
freemanzMrojo, line 213: Suggestion: Talking about Russian dolls, this block of code looks also pretty similar to the above where I put the comment at `node.go`

chain/repository.go
vanja-vechain: Sure will apply suggestion, reason for storing them is that we keep cache of evidences as we see double signing on the network. One is applied, so in next block next one will be applied, if it is not cached we would have to go and check for each block was there any double signing in previous X blocks.

cmd/thor/node/node.go
vanja-vechain: So, here are two separate functionalities, first block of code L357-365 is recording if the new block has double signed sibling and records this. Then, lines 377-384 are checking did proposer add evidence about previous double signing, which happened in the past, if yes then it validates those evidences as part of block validation. And last one 451-459 is actually removing verified evidences from cache. So 1 block is totally separate functionality from blocks 2 and 3 and gap between 2 and 3 is because between there is consensus check so we don't want to remove from cache until we are sure block is ok. If that makes sense?

cmd/thor/node/packer_loop.go
vanja-vechain: This cannot be moved since it has to mark the evidence as processed only after block is broadcasted to network

cmd/thor/node/node.go
freemanzMrojo: yep, makes sense, thanks for clarifying! maybe a comment or two would help, I struggled to understand the reasoning 😢

cmd/thor/node/packer_loop.go
freemanzMrojo: I meant the code in a method and that method called from here, not removed. Sorry for the misunderstanding 👍🏼

cmd/thor/node/node.go
vanja-vechain: Refactored to headers

chain/repository.go
vanja-vechain: Applied suggestion

packer/flow.go
vanja-vechain: Fixed, cheers!

YeahNotSewerSide (approved):
lgtm

PR #1161, opened 2025-07-10
refactor(staker): configure gas usage at solidity level

darrenvechain:
# Description

Configure gas usage at the `solidity` package level

# Checklist:

### DIFF
--- a/builtin/solidity/bytes32.go
+++ b/builtin/solidity/bytes32.go
@@ -6,28 +6,32 @@
 package solidity
 
 import (
-	"github.com/vechain/thor/v2/state"
 	"github.com/vechain/thor/v2/thor"
 )
 
 // Bytes32 is a wrapper for storage and retrieval of [32]byte
 type Bytes32 struct {
-	contract thor.Address
-	state    *state.State
-	pos      thor.Bytes32
+	context *Context
+	pos     thor.Bytes32
 }
 
-func NewBytes32(contract thor.Address, state *state.State, pos thor.Bytes32) *Bytes32 {
-	return &Bytes32{contract: contract, state: state, pos: pos}
+func NewBytes32(context *Context, pos thor.Bytes32) *Bytes32 {
+	return &Bytes32{context: context, pos: pos}
 }
 
 func (a *Bytes32) Get() (thor.Bytes32, error) {
-	return a.state.GetStorage(a.contract, a.pos)
+	a.context.UseGas(thor.SloadGas)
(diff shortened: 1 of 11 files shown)


YeahNotSewerSide (approved):

builtin/gascharger/gas_charger.go
```diff
@@ -64,7 +64,9 @@ func (c *Charger) Charge(gas uint64) {
 		c.customGas += gas
 	}
 
-	c.env.UseGas(gas)
+	if c.env != nil {
+		c.env.UseGas(gas)
+	}
```
otherview: This shouldn't be possible to happen, right ? I understand the concern, make sense to guarantee the env is not nil.
However doing it here would mean silently avoiding gas usage. What if this check was done in the `New()` ?

builtin/solidity/uint256_test.go
```diff
@@ -0,0 +1,57 @@
...
+	charger = gascharger.New(nil)
+	ctx.charger = charger
+
+	err = uint.Sub(big.NewInt(200))
+	assert.NoError(t, err)
+	assert.Equal(t, ctx.charger.TotalGas(), thor.SstoreResetGas+thor.SloadGas)
+
+	value, err = uint.Get()
+	assert.NoError(t, err)
+	assert.Equal(t, big.NewInt(1300), value)
```
otherview, line 56: Worth adding an error check if the subbed value totals -1 ? 

Also, do you think we should check for overflows ?

otherview (approved):
lgtm - minor comments, no blocker to merge imo

builtin/gascharger/gas_charger.go
darrenkel: Ya the purpose here was for the unit tests in the `solidity` package:

```go
	charger := gascharger.New(nil)  
	mapping.context.charger = charger // Reset charger to measure only the SLOAD operation
```

Maybe should have a `Reset` function on the gas charger?

otherview: Makes sense, I think a charger with nil should panic ( doesnt do anything) and would help avoid any miss step when hooking it.

For the test, would the method `SetTestHook` work ?

builtin/solidity/uint256_test.go
darrenkel: Added a check/ test for both

builtin/gascharger/gas_charger.go
darrenkel: Pushed a change, to create the xenv within the unit tests, was easy enough

PR #1171, opened 2025-07-15
fix(staker): various increase/decrease/renew test cases

darrenvechain:
# Description

# Checklist:

### DIFF
--- a/builtin/staker/delegations.go
+++ b/builtin/staker/delegations.go
@@ -60,9 +60,9 @@ func (d *delegations) Add(
 	if aggregated.IsEmpty() {
 		aggregated = newAggregation()
 	}
-	nextPeriodStake := validation.NextPeriodStakes(aggregated)
-	nextPeriodStake = nextPeriodStake.Add(nextPeriodStake, stake)
-	if nextPeriodStake.Cmp(MaxStake) > 0 {
+	nextPeriodTVL := big.NewInt(0).Add(validation.NextPeriodTVL(), aggregated.NextPeriodTVL())
+	nextPeriodTVL = nextPeriodTVL.Add(nextPeriodTVL, stake)
+	if nextPeriodTVL.Cmp(MaxStake) > 0 {
 		return thor.Bytes32{}, errors.New("validation's next period stake exceeds max stake")
 	}
 
@@ -84,10 +84,13 @@ func (d *delegations) Add(
 		ValidationID:   validationID,
 		FirstIteration: validation.CurrentIteration() + 1,
 	}
-
 	weight := delegation.Weight()
-
-	if !autoRenew {
+	if delegation.AutoRenew {
+		aggregated.PendingRecurringVET = big.NewInt(0).Add(aggregated.PendingRecurringVET, delegation.Stake)
+		aggregated.PendingRecurringWeight = big.NewInt(0).Add(aggregated.PendingRecurringWeight, weight)
+	} else {
+		aggregated.PendingOneTimeVET = big.NewInt(0).Add(aggregated.PendingOneTimeVET, delegation.Stake)
+		aggregated.PendingOneTimeWeight = big.NewInt(0).Add(aggregated.PendingOneTimeWeight, weight)
 		last := validation.CurrentIteration() + 1
 		delegation.LastIteration = &last
 	}
@@ -95,19 +98,9 @@ func (d *delegations) Add(
 	if err := d.queuedVET.Add(stake); err != nil {
 		return thor.Bytes32{}, err
 	}
-
 	if err := d.queuedWeight.Add(weight); err != nil {
 		return thor.Bytes32{}, err
 	}
-
-	if delegation.AutoRenew {
-		aggregated.PendingRecurringVET = big.NewInt(0).Add(aggregated.PendingRecurringVET, delegation.Stake)
-		aggregated.PendingRecurringWeight = big.NewInt(0).Add(aggregated.PendingRecurringWeight, weight)
-	} else {
-		aggregated.PendingOneTimeVET = big.NewInt(0).Add(aggregated.PendingOneTimeVET, delegation.Stake)
-		aggregated.PendingOneTimeWeight = big.NewInt(0).Add(aggregated.PendingOneTimeWeight, weight)
-	}
-
 	if err := d.storage.SetAggregation(validationID, aggregated, false); err != nil {
 		return thor.Bytes32{}, err
 	}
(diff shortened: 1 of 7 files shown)


builtin/staker/housekeeping.go
```diff
@@ -112,7 +112,7 @@ func (s *Staker) performRenewalUpdates(id thor.Bytes32, validator *Validation) e
 	queuedWeight := big.NewInt(0).Add(validatorRenewal.QueuedDecreaseWeight, delegationsRenewal.QueuedDecreaseWeight)
 
 	// set the new totals
-	validator.LockedVET = big.NewInt(0).Add(validator.LockedVET, changeTVL)
+	validator.LockedVET = big.NewInt(0).Add(validator.LockedVET, validatorRenewal.ChangeTVL)
```
darrenkel, line 115: severe bug

builtin/staker/staker.go
```diff
@@ -352,8 +352,9 @@ func (s *Staker) GetValidatorsTotals(validationID thor.Bytes32) (*ValidationTota
 	if err != nil {
 		return nil, err
 	}
+	delegationLockedStake := big.NewInt(0).Add(aggregation.CurrentRecurringVET, aggregation.CurrentOneTimeVET)
```
YeahNotSewerSide, line 355: do we need to return sum of total stake(DelegationsLockedStake+ValidationLockedVet) if we also have `DelegationsLockedStake`?

darrenkel: Ya I had tried returning just the validator locked VET but it broke e2e, so I decided to make it backwards compatible

builtin/staker/delegations.go
```diff
@@ -84,30 +84,23 @@ func (d *delegations) Add(
...
 	weight := delegation.Weight()
-
-	if !autoRenew {
+	if delegation.AutoRenew {
+		aggregated.PendingRecurringVET = big.NewInt(0).Add(aggregated.PendingRecurringVET, delegation.Stake)
+		aggregated.PendingRecurringWeight = big.NewInt(0).Add(aggregated.PendingRecurringWeight, weight)
+	} else {
+		aggregated.PendingOneTimeVET = big.NewInt(0).Add(aggregated.PendingOneTimeVET, delegation.Stake)
+		aggregated.PendingOneTimeWeight = big.NewInt(0).Add(aggregated.PendingOneTimeWeight, weight)
 		last := validation.CurrentIteration() + 1
```
freemanzMrojo, line 94: Just a question so I understand the reasoning behind the changes.

Before we had these 2 lines when there was no `autoRenew`. Now we have them only in the case the autorenew is enabled for the delegation.

I am not sure whether the `autoRenew` parameter in this method applies to validators only (and `delegation.AutoRenew` just to their delegations). Is that the case? If not, how should I understand the change?

builtin/staker/delegations.go
```diff
@@ -166,9 +159,18 @@ func (d *delegations) EnableAutoRenew(delegationID thor.Bytes32) error {
...
 	}
-
+	if delegation.AutoRenew {
+		return errors.New("delegation is already autoRenew")
+	}
 	weight := delegation.Weight()
 
+	// validate that the enablement does not exceed the max stake considering next staking period changes
+	nextPeriodTVL := big.NewInt(0).Add(validation.NextPeriodTVL(), aggregation.NextPeriodTVL())
+	nextPeriodTVL.Add(nextPeriodTVL, delegation.Stake)
```
freemanzMrojo, line 169: Suggestion: Not sure whether the `stake` in the combination above is also a delegation stake but in that case it might make sense to create a method so we just run these 2 lines (validation tvl + aggregation tvl + delegation stake)

builtin/staker/housekeeping.go
```diff
@@ -112,7 +112,7 @@ func (s *Staker) performRenewalUpdates(id thor.Bytes32, validator *Validation) e
...
 	validator.Weight = big.NewInt(0).Add(validator.Weight, changeWeight)
```
freemanzMrojo, line 116: Question for my own understanding: If the above is a severe bug, why this line is not `validator.Weight = big.NewInt(0).Add(validator.Weight, validatorRenewal.ChangeWeight)` instead too?

builtin/staker/types.go
```diff
@@ -31,7 +31,7 @@ type Validation struct {
 	StartBlock         uint32       // the block number when the validation started the first staking period
 	ExitBlock          *uint32      `rlp:"nil"` // the block number when the validation moved to cooldown
 
-	LockedVET          *big.Int // the amount of VET locked for the current staking period
+	LockedVET          *big.Int // the amount of VET locked for the current staking period, for the validator only
```
freemanzMrojo, line 34: One thing that name-wise confuses me. If this applies to the validator only, should not we modify either the method [here](https://github.com/vechain/thor/blob/release/hayabusa/builtin/staker/staker.go#L79) or this attribute? `staker.LockedVET` returns (according to the comment) stake and weight including delegators, as far I understand

builtin/staker/delegations.go
darrenkel: Ya previously we have 2 ifs for `if autorenew`, I compressed the code to use the same if block.

For the 2nd part, the `autoRenew` applies only to delegations, not validations.

If the validation has auto renew off and then exits, the delegation will exit as well by default

builtin/staker/delegations.go
darrenkel: Since we asserted that the delegation was not previously in auto renew, we know that `aggregation.NextPeriodTVL()` does not contain this stake.

`NextPeriodTVL` returns the expected stake in the next staking period, so does not consider the current delegation to be part of that (since it was previously autorenew = false)

builtin/staker/housekeeping.go
darrenkel: So weight is stored in the validation as the combined value for itself and all delegations. The reason being for the scheduler we don't want to do extra reads from state / additions etc. 

Also, a validator cannot withdraw their weight, only their locked vet. The comments on the `Validation` struct may help.

The `LockedVET` field is intended to be only for the individual validation.

builtin/staker/types.go
darrenkel: Yepp, good spot, that method should be updated.

darrenkel: Ah sorry, `LockedVET` function does not look at individual validations/ delegations.

Its a contract total, so should be okay I think

builtin/staker/housekeeping.go
freemanzMrojo: Ok, thanks

builtin/staker/delegations.go
freemanzMrojo: Perfect, thanks

builtin/staker/delegations.go
freemanzMrojo: Understood, thanks.

builtin/staker/types.go
freemanzMrojo: Ok, thanks

freemanzMrojo: Btw, I saw your last comment late. What I meant is that:

- Individual LockedVET (the above) => As per this PR, only for the validators
- Total LockedVET (`staker.LockedVET`) => Validators + their delegations

Hence my confusion.

YeahNotSewerSide (approved):

PR #1183, opened 2025-07-22
feat: finality based on weight

freemanzMrojo:
# Description

Implements finality based on weight. Tests adapted so they run using both the previous fork config and Hayabusa.


paologalligit (approved):
Looks good! 👍

bft/engine.go
```diff
@@ -317,17 +317,38 @@ func (engine *Engine) computeState(header *block.Header) (*bftState, error) {
 	h := header
 	for h.Number() >= engine.forkConfig.FINALITY {
 		signer, _ := h.Signer()
-		js.AddBlock(signer, h.COM())
+
+		var parentBlockSummary *chain.BlockSummary
+		var err error
+
+		if h.Number() > engine.forkConfig.HAYABUSA+engine.forkConfig.HAYABUSA_TP {
```
vanja-vechain, line 324: There is an edge case here, hayabusa doesn't have to switch at this point (if there are not enough validators). IsPoSActive() on staker is safer to use

bft/justifier.go
```diff
@@ -58,33 +73,63 @@ func (engine *Engine) newJustifier(parentID thor.Bytes32) (*justifier, error) {
...
 	}
 
-	return &justifier{
-		votes:         make(map[thor.Address]bool),
-		parentQuality: parentQuality,
-		checkpoint:    checkpoint,
-		threshold:     threshold,
-	}, nil
+	// When the checkpoint is greater than 0, last of parent round is checkpoint - 1
+	if lastOfParentRound >= engine.forkConfig.HAYABUSA+engine.forkConfig.HAYABUSA_TP {
```
vanja-vechain: same here

bft/engine.go
freemanzMrojo: We need to check against the block number, and `IsPoSActive` only checks against the staker without using a number as far as I see. You mean to also check that here?

bft/justifier.go
freemanzMrojo: We need to check against the block number, and IsPoSActive only checks against the staker without using a number as far as I see. You mean to also check that here?

The checkpoint might be in the past.

bft/engine.go
vanja-vechain: Yes, since engine.forkConfig.HAYABUSA+engine.forkConfig.HAYABUSA_TP doesn't have to always be a point of transition

freemanzMrojo: But this is a loop that go backwards. So if we check against `staker.IsPosActive` it might be wrongly true because in the present it is active but not when the block in the check happened.

freemanzMrojo: I can think of a solution though @vanja-vechain so IsPoSActive works (by instantiating from the state every block):

```
		state := engine.stater.NewState(parentBlockSummary.Root())
		staker := builtin.Staker.Native(state)

		hayabusaActive, err := staker.IsPoSActive()
		if err != nil {
			return nil, err
		}
```

wdyt?

freemanzMrojo: Done, thanks

bft/justifier.go
freemanzMrojo: Done, thanks

vanja-vechain (approved):

PR #1185, opened 2025-07-23
feat(solo): instantly mine on POST /transactions

darrenkel:
# Description

Instantly mine on POST /transactions, uses existing `on-demand` flag instead of creating a new flag

image

# Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,4 +104,3 @@ A special shout out to following projects:
 
 VeChainThor is licensed under the [GNU Lesser General Public License v3.0](https://www.gnu.org/licenses/lgpl-3.0.html),
 also included in *LICENSE* file in repository.
-


cmd/thor/main.go
```diff
@@ -422,24 +422,45 @@ func soloAction(ctx *cli.Context) error {
...
+		SkipLogs:         skipLogs,
+		MinTxPriorityFee: minTxPriorityFee,
+		OnDemand:         onDemandBlockProduction,
+		BlockInterval:    blockProductionInterval,
 	}
 
-	txPool := txpool.New(repo, state.NewStater(mainDB), txPoolOption, forkConfig)
-	defer func() { log.Info("closing tx pool..."); txPool.Close() }()
+	stater := state.NewStater(mainDB)
+	engine := solo.NewEngine(repo, stater, logDB, options, forkConfig)
```
libotony: Can you move `engine` initialization to `solo.New`? It is confused with `bft.Engine`.

cmd/thor/solo/txpool.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (o *OnDemandTxPool) AddLocal(newTx *tx.Transaction) error {
+	o.mu.Lock()
+	defer o.mu.Unlock()
+
+	o.txsByHash[newTx.Hash()] = newTx
+	o.txsByID[newTx.ID()] = newTx
+
+	executable, err := o.engine.IsExecutable(newTx)
```
libotony, line 53: Checks in `IsExecutable` will be performed by the `o.engine.Pack`.

cmd/thor/solo/txpool.go
```diff
@@ -0,0 +1,129 @@
...
+	"github.com/vechain/thor/v2/txpool"
+)
+
+type OnDemandTxPool struct {
+	engine *Engine
+
+	txsByHash map[thor.Bytes32]*tx.Transaction
+	txsByID   map[thor.Bytes32]*tx.Transaction
+
+	txFeed event.Feed
```
libotony, line 24: `TxFeed` can be removed, no usage if you don't use it in solo package

libotony: And you can keep only one of  `ByHash` or `ByID`

cmd/thor/solo/engine.go
```diff
@@ -0,0 +1,158 @@
...
+	"github.com/vechain/thor/v2/genesis"
+	"github.com/vechain/thor/v2/logdb"
+	"github.com/vechain/thor/v2/packer"
+	"github.com/vechain/thor/v2/state"
+	"github.com/vechain/thor/v2/thor"
+	"github.com/vechain/thor/v2/tx"
+	"github.com/vechain/thor/v2/txpool"
+)
+
+type Engine struct {
```
libotony: Nit; maybe `Core` but Engine is fine.

cmd/thor/main.go
darrenkel: We need to wrap both the pool and the solo loop around the `Pack` method, so needed a separate struct I think

cmd/thor/solo/txpool.go
darrenkel: > `TxFeed` can be removed, no usage if you don't use it in solo package

Its used by the tx pool subscription (API)

cmd/thor/solo/txpool.go
darrenkel: The tx gets instantly removed if we don't do this I think. This allows the API to add txs that are not currently executable

Previously this behaviour was handled by the real tx pool, so I was trying to mimic the behaviour

cmd/thor/main.go
darrenkel: Ah, think its just naming convention you mean? I updated to `Core`

cmd/thor/solo/txpool.go
libotony: It's a bit complicated, if you don't mind I will make a followup PR after this is merged

darrenkel: Feel free to target a PR on this branch if you like

libotony: Solved offline.

PR #1192, opened 2025-07-24
Refactoring staker 

otherview:
# Description

This PR refactors the staker package architecture to improve maintainability and separation of concerns by extracting delegation aggregations and global statistics into dedicated services.

### Key Changes

**Service Extraction:**
- Created `aggregation.Service` to manage per-validator delegation totals (locked, pending, exiting, withdrawable VET)
- Created `globalstats.Service` to manage contract-wide staking statistics
- Each service manages its own storage slots and encapsulates related logic

**Architecture Improvements:**
- Introduced `delta` package with `Renewal` and `Exit` types for clean state transitions
- Decoupled aggregation calculations from core validation/delegation logic
- Main `Staker` struct now orchestrates between services rather than handling everything directly

**Benefits:**
- **Better separation of concerns**: Each service has a single, well-defined responsibility
- **Enhanced maintainability**: Related functionality is grouped together, reducing coupling
- **Cleaner state transitions**: Delta objects provide explicit tracking of changes during period renewals

### Technical Details

- `aggregation.Service` handles per-validator delegation totals with separate tracking for different VET states
- `globalstats.Service` maintains contract-wide locked/queued VET and weight totals
- Period transitions now use delta objects to communicate changes between services
- Storage access is encapsulated within each service, reducing direct storage manipulation

# Checklist:


builtin/staker/validations.go
```diff
@@ -418,50 +318,38 @@ func (v *validations) GetWithdrawable(id thor.Address, currentBlock uint32) (*Va
...
 	entry.CooldownVET = big.NewInt(0).Set(entry.LockedVET)
 	entry.LockedVET = big.NewInt(0)
 	entry.NextPeriodDecrease = big.NewInt(0)
-	// unlock delegator's stakes and remove their weight
-	entry.CompleteIterations++
+	entry.Weight = big.NewInt(0)
 
+	// unlock pending stake
 	if entry.PendingLocked.Sign() == 1 {
+		onCoolDownTVL = big.NewInt(0).Add(onCoolDownTVL, entry.PendingLocked)
```
darrenkel: Curious, why do we add it to onCoolDownTVL?

builtin/staker/housekeeping.go
```diff
@@ -93,41 +106,27 @@ func (s *Staker) Housekeep(currentBlock uint32) (bool, map[thor.Address]*Validat
...
+	delegationsRenewal, err := s.aggregationService.Renew(validationID)
 	if err != nil {
 		return err
 	}
 
-	// renew the validator & delegations
-	validatorRenewal := validator.Renew()
-	delegationsRenewal := aggregation.Renew()
-	if err := s.storage.SetAggregation(id, aggregation, false); err != nil {
+	if err = s.globalStatsService.UpdateTotals(validatorRenewal, delegationsRenewal); err != nil {
```
darrenkel, line 121: FYI @vanja-vechain maybe could apply your change here.

This code writes to each total 101 times, better to accumulate and write to each total once

builtin/staker/aggregation/aggregation.go
```diff
@@ -0,0 +1,131 @@
...
+	// Return these values to modify contract totals
+	exitedTVL := big.NewInt(0).Set(a.LockedVET)
+	exitedWeight := big.NewInt(0).Set(a.LockedWeight)
+	queuedDecrease := big.NewInt(0).Set(a.PendingVET)
+	queuedWeightDecrease := big.NewInt(0).Set(a.PendingWeight)
+
+	// Move all the funds to withdrawable
+	withdrawable := big.NewInt(0).Set(a.WithdrawableVET)
+	withdrawable.Add(withdrawable, a.LockedVET)
+	withdrawable.Add(withdrawable, a.PendingVET)
```
vanja-vechain, line 114: Should we add a.ExitingVet here as well, example before the validator signaled exit, some delegator signaled exit before, his stake would be in exiting vet in that case, if I'm not wrong

builtin/staker/aggregation/service.go
```diff
@@ -0,0 +1,140 @@
...
+// GetAggregation retrieves the delegation aggregation for a validator.
+// Returns a zero-initialized aggregation if none exists.
+func (s *Service) GetAggregation(validationID thor.Address) (*Aggregation, error) {
+	d, err := s.aggregationStorage.Get(validationID)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get validator aggregation")
+	}
+
+	// never return nil pointer aggregations
+	// should never happen a case where d.LockedVET == nil and d.WithdrawableVET != nil
```
vanja-vechain: based on the comment here, should it be if d == nil || d.LockedVET == nil  || d.WithdrawableVET != nil or maybe even better we could use d.IsEmpty()

builtin/staker/validations.go
otherview: Good catch, I can't see the reason for it - would assume that was something I was debugging, should have cleaned that up.
Will add a todo to create a test for this scenario.

builtin/staker/aggregation/aggregation.go
darrenkel: @vanja-vechain `ExitingVET` does not represent  any stored VET, its an indication to decrease `LockedVET` on next staking renewal. Since they are not renewing, its safe to set to 0

vanja-vechain: Ok, ignore above, now i see that it remains in the locked, in agg service

builtin/staker/globalstats/service.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+	if err := s.queuedWeight.Sub(queuedWeight); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// GetLocketVET returns the total VET and weight currently locked in active staking.
+func (s *Service) GetLocketVET() (*big.Int, *big.Int, error) {
```
vanja-vechain: minor Locked instead of Locket

builtin/staker/housekeeping.go
vanja-vechain: Sure, will do once merged

vanja-vechain (approved):

builtin/staker/aggregation/aggregation.go
otherview: I think that case is handled in the renew() - when the delegator signals exit it's it will eventually renew(), when it does deducts all exitingVET from locked and adds it to the withdrawal.

So when the exit function is called there should be no exiting vet as it should have been handled in the renewal.

Will add a comment to make this clear 🙏 

```
// Remove ExitingVET from LockedVET
	changeTVL = changeTVL.Sub(changeTVL, a.ExitingVET)
	changeWeight = changeWeight.Sub(changeWeight, a.ExitingWeight)
	a.LockedVET = big.NewInt(0).Sub(a.LockedVET, a.ExitingVET)
	a.LockedWeight = big.NewInt(0).Sub(a.LockedWeight, a.ExitingWeight)
	
// Move ExitingVET to WithdrawableVET
	a.WithdrawableVET = big.NewInt(0).Add(a.WithdrawableVET, a.ExitingVET)
	a.ExitingVET = big.NewInt(0)
	a.ExitingWeight = big.NewInt(0)
	```

File: builtin/staker/globalstats/service.go [hunk 6]
```diff
@@ -0,0 +1,144 @@
...
+		return err
+	}
+	// unlockedTVLWeight already has the sum of the agg weights - LockedVET x2 + total weight from delegators
+	if err := s.lockedWeight.Sub(unlockedTVLWeight); err != nil {
+		return err
+	}
+	if err := s.queuedVET.Sub(aggExit.QueuedDecrease); err != nil {
+		return err
+	}
+	if err := s.queuedWeight.Sub(aggExit.QueuedDecreaseWeight); err != nil {
```
darrenkel, line 132: Can the validator have a queued amount as well?

builtin/staker/globalstats/service.go
```diff
@@ -0,0 +1,144 @@
...
+	totalUnlockedVET := big.NewInt(0).Add(unlockedTVL, aggExit.ExitedTVL)
+
+	if err := s.lockedVET.Sub(totalUnlockedVET); err != nil {
+		return err
+	}
+	// unlockedTVLWeight already has the sum of the agg weights - LockedVET x2 + total weight from delegators
+	if err := s.lockedWeight.Sub(unlockedTVLWeight); err != nil {
+		return err
+	}
+	if err := s.queuedVET.Sub(aggExit.QueuedDecrease); err != nil {
```
darrenkel: Same as below, can this apply to validators

builtin/staker/aggregation/service.go
otherview: d.IsEmpty() might return and NPE if, for example d.WithdrawableVET == nil.

The comment there means to say that checking for a single object to be nil is enough because either d = nil or d.property = nil. No need to check all d.properties.
It can never happen a case where d.property1 = nil and d.property2 != nil.

I'll update the comment

builtin/staker/globalstats/service.go
otherview: engrish fixed !

builtin/staker/globalstats/service.go
otherview: Yes I believe so. Will add a todo to add a test for that scenario 👍

otherview: Wait, that is the reason for https://github.com/vechain/thor/pull/1192#discussion_r2231161760 ! 

Validators queued amount is pending locked. which is added to the withdrawable upon exit 😅 
I think using onCoolDownTVL is misleading, will change that

vanja-vechain (approved):

builtin/staker/globalstats/service.go
otherview: Point still remains that they should be added to the decrease of the queued vet. fixed 👍

builtin/staker/globalstats/service.go
otherview: Added a todo to impl a test 👍

libotony (approved):

PR #1206, opened 2025-07-29
Decouple calculation and application in housekeep

otherview:
# Description

This PR decouples the calculations and application of new state in the housekeep function.

# Checklist:

### DIFF
--- a/builtin/staker/housekeep.go
+++ b/builtin/staker/housekeep.go
@@ -80,7 +80,7 @@ func (s *Staker) ComputeEpochTransition(currentBlock uint32) (*EpochTransition, 
 	}
 
 	// 3. Compute all activations
-	transition.ActivationCount, err = s.computeActivations()
+	transition.ActivationCount, err = s.computeActivations(transition.ExitValidatorID != nil)
 	if err != nil {
 		return nil, err
 	}
@@ -156,7 +156,7 @@ func (s *Staker) computeExits(currentBlock uint32) (*thor.Address, error) {
 	return nil, nil
 }
 
-func (s *Staker) computeActivations() (int64, error) {
+func (s *Staker) computeActivations(hasValidatorExited bool) (int64, error) {
 	// Calculate how many validators can be activated
 	queuedSize, err := s.QueuedGroupSize()
 	if err != nil {
@@ -166,6 +166,11 @@ func (s *Staker) computeActivations() (int64, error) {
 	if err != nil {
 		return 0, err
 	}
+	// the current leaderSize might have changed for the next activations
+	if hasValidatorExited {
+		leaderSize = big.NewInt(0).Sub(leaderSize, big.NewInt(1))
+	}
+
 	maxSize, err := s.params.Get(thor.KeyMaxBlockProposers)
 	if err != nil {
 		return 0, err


builtin/staker/housekeep.go
```diff
@@ -0,0 +1,429 @@
...
+
+// Housekeep performs epoch transitions at epoch boundaries
+func (s *Staker) Housekeep(currentBlock uint32) (bool, map[thor.Address]*validation.Validation, error) {
+	if currentBlock%epochLength != 0 {
+		return false, nil, nil
+	}
+
+	logger.Info("🏠performing housekeeping", "block", currentBlock)
+
+	transition, err := s.ComputeEpochTransition(currentBlock)
```
darrenkel, line 41: Looks like we're doing quite a lot more reads for these computations. Should we bench mark them?, eg. I see  3x `LeaderGroupIterator`, where we used to have 1

builtin/staker/housekeep.go
```diff
@@ -0,0 +1,429 @@
...
+}
+
+// ApplyEpochTransition applies all computed changes
+func (s *Staker) ApplyEpochTransition(transition *EpochTransition) error {
+	logger.Info("applying epoch transition", "block", transition.Block)
+
+	// Apply renewals
+	for _, renewal := range transition.Renewals {
+		// Update global stats with existing service methods
+		if err := s.globalStatsService.UpdateTotals(renewal.ValidatorDelta, renewal.DelegationDelta); err != nil {
```
darrenkel, line 204: We should accumulate these renewals I think? Call `UpdateTotals` with the accumulated to reduce writes to storage

darrenkel: Maybe this hasn't been merged already, I think Vanja opened a PR for it

builtin/staker/housekeep.go
otherview: Yep, definitely worth doing. Don't want to spin unnecessarily. But where do you see the considerable more reads ?

darrenkel: We do `LeaderGroupIterator` 3 times, we used to do it once

darrenkel: How about call the LeaderGroup() function and perform computations off the common map

otherview: Yeah, there is certainly a trade-off to make the ops more visible.
I think if the storage has a cache, it shouldn't be an issue.

But perhaps I can modify the `computeRenewals` and the `computeExits` to use one cycle

builtin/staker/housekeep.go
otherview: It's a good point, added a todo so to not forget it.

builtin/staker/housekeep.go
darrenkel: Not sure we want to have a cache here, it will prob lead to more bugs.

otherview: I would typically agree, but because it's a single thread centralized storage environment, I think should be pretty safe. 

Yeah, it's worth having a baseline first so we can see what's the hit first.

darrenkel: Looks like its very quick to read the leadergroup, so looks okay. 1.2 seconds to call `LeaderGroupIterator` over 7k times

image

darrenkel:
```go
func BenchmarkStaker_LeaderGroup(b *testing.B) {
	b.Logf("Benchmarking %d stakers", b.N)

	stakers := make([]*Staker, b.N)
	for i := 0; i < b.N; i++ {
		staker, _ := newStaker(b, 101, 101, true)
		stakers[i] = staker
	}

	b.Logf("Created %d stakers in %s", b.N, b.Elapsed())
	b.ResetTimer() // Reset the timer to exclude creation time from the benchmark

	for i := 0; i < b.N; i++ {
		staker := stakers[i]
		_, err := staker.LeaderGroup()
		assert.NoError(b, err)
	}
	b.Logf("Processed %d stakers in %s", b.N, b.Elapsed())
}
```

otherview: Yeah, it should be ok given this is processed once every epoch. But the overhead might be visible when processing blocks. 

I'll create a ticket to review this.

otherview: Created ticket in https://github.com/orgs/vechain/projects/22/views/14

PR #1212, opened 2025-07-30
Refactor Transition

otherview:
# Description

Please include a summary of the changes and the related issue. Please also include relevant motivation and context. List
any dependencies that are required for this change.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:

### DIFF
--- a/builtin/staker/transition.go
+++ b/builtin/staker/transition.go
@@ -1,0 +1,94 @@
+package staker
+
+import (
+	"github.com/vechain/thor/v2/thor"
+	"math/big"
+)
+
+// Transition activates the staker contract when sufficient validators are queued
+func (s *Staker) Transition(currentBlock uint32) (bool, error) {
+	active, err := s.IsPoSActive()
+	if err != nil {
+		return false, err
+	}
+	if active {
+		return false, nil
+	}
+
+	maxProposers, err := s.params.Get(thor.KeyMaxBlockProposers)
+	if err != nil || maxProposers.Cmp(big.NewInt(0)) == 0 {
+		maxProposers = big.NewInt(0).SetUint64(thor.InitialMaxBlockProposers)
+	}
+
+	queueSize, err := s.validationService.QueuedGroupSize()
+	if err != nil {
+		return false, err
+	}
+
+	// if the queue size is not AT LEAST 2/3 of the maxProposers, then return nil
+	minimum := big.NewFloat(0).SetInt(maxProposers)
+	minimum.Mul(minimum, big.NewFloat(2))
+	minimum.Quo(minimum, big.NewFloat(3))
+	if big.NewFloat(0).SetInt(queueSize).Cmp(minimum) < 0 {
+		return false, nil
+	}
+
+	// Use existing activateValidators method for transition
+	ids, err := s.activateValidators(currentBlock)
+	if err != nil {
+		return false, err
+	}
+	logger.Info("activated validations", "count", len(ids))
+
+	return true, nil
+}
+
+// activateValidators is kept for the Transition method
+func (s *Staker) activateValidators(currentBlock uint32) ([]*thor.Address, error) {
+	queuedSize, err := s.QueuedGroupSize()
+	if err != nil {
+		return nil, err
+	}
+	leaderSize, err := s.LeaderGroupSize()
+	if err != nil {
+		return nil, err
+	}
+	maxSize, err := s.params.Get(thor.KeyMaxBlockProposers)
+	if err != nil {
+		return nil, err
+	}
+	if leaderSize.Cmp(maxSize) >= 0 {
+		return nil, nil
+	}
+
+	// no one is in the queue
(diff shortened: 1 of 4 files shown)


builtin/staker/validation/service.go
```diff
@@ -424,6 +421,59 @@ func (s *Service) GetDelegatorRewards(validationID thor.Address, stakingPeriod u
...
+	added, err := s.leaderGroup.Add(validationID, val)
+	if err != nil {
+		return nil, err
+	}
+	if !added {
+		return nil, errors.New("failed to add validator to active list")
+	}
+
+	// Persist the updated validation state
+	if err = s.SetValidation(validationID, val, false); err != nil {
```
darrenkel, line 466: Think validator is already added above:

```
added, err := s.leaderGroup.Add(validationID, val)
```

Can either remove this or we should probably remove write ops in the linked list

otherview: Yeap, good catch, I had noticed that - meant to review the linked list next.

Let me try to clean up that double storing 🙏  Should be ok to just remove that behaviour from the LL

otherview: Yeah, same thing happens in the Remove ops. Adding TODO's there 👍

builtin/staker/validation/service.go
```diff
@@ -424,6 +421,59 @@ func (s *Service) GetDelegatorRewards(validator thor.Address, stakingPeriod uint
...
+	currentBlock uint32,
+	aggRenew *delta.Renewal,
+) (*delta.Renewal, error) {
+	val, err := s.GetExistingValidation(validationID)
+	if err != nil {
+		return nil, err
+	}
+
+	// Update validator values
+	validatorLocked := big.NewInt(0).Add(val.LockedVET, val.QueuedVET) // lock exiting + pending vet
```
paologalligit:
```suggestion
	validatorLocked := big.NewInt(0).Add(val.LockedVET, val.QueuedVET) // lock existing + pending vet
```

builtin/staker/validation/service.go
```diff
@@ -424,6 +421,59 @@ func (s *Service) GetDelegatorRewards(validator thor.Address, stakingPeriod uint
...
+	// Persist the updated validation state
+	if err = s.SetValidation(validationID, val, false); err != nil {
+		return nil, err
+	}
+
+	// Return delta representing the state changes
+	validatorRenewal := &delta.Renewal{
+		NewLockedVET:         val.LockedVET,
+		NewLockedWeight:      val.Weight,
+		QueuedDecrease:       val.LockedVET,
```
paologalligit, line 474: Is this correct? From the name I'm expecting this to be the amount that was subtracted by the `queuedVET`, not the `newLockedVET`. Same the `QueuedDecreaseWeight` below

builtin/staker/transition.go
```diff
@@ -0,0 +1,64 @@
...
+	active, err := s.IsPoSActive()
+	if err != nil {
+		return false, err
+	}
+	if active {
+		return false, nil
+	}
+
+	maxProposers, err := s.params.Get(thor.KeyMaxBlockProposers)
+	if err != nil || maxProposers.Cmp(big.NewInt(0)) == 0 {
```
paologalligit: Is it safe here to treat the error and empty case in the same way? Just curious, probably in this case it is, but wanted to double check

builtin/staker/transition.go
```diff
@@ -0,0 +1,64 @@
...
+		maxProposers = big.NewInt(0).SetUint64(thor.InitialMaxBlockProposers)
+	}
+
+	queueSize, err := s.validationService.QueuedGroupSize()
+	if err != nil {
+		return false, err
+	}
+
+	// Transitions is not possible if the queue size is not AT LEAST 2/3 of the maxProposers
+	minimum := big.NewFloat(0).SetInt(maxProposers)
```
paologalligit: What's the reason behind using float instead of int here? So we don't need to round up?

builtin/staker/transition.go
otherview: Yeah, worth surfacing these for sure 👍

builtin/staker/transition.go
otherview: Nice one, found a good simplification for this one.
```
	// Transitions is not possible if the queue size is not AT LEAST 2/3 of the maxProposers
	// queueSize >= 2/3 * maxProposers
	// queueSize * 3 >= maxProposers * 2
	if queueSize.Int64()*3 < maxProposers.Int64()*2 {
		return false, nil
	}
```

builtin/staker/validation/service.go
otherview: Good catch - The validator should never have locked VET until it's activated.

paologalligit (approved):

PR #1213, opened 2025-07-31
chore: init staker params once

darrenvechain:
# Description

- This reduces reads to state. The params only get debugged once, instead of every time the staker is initialised
- It also ensures params cannot be set by an outside package

# Checklist:

### DIFF
--- a/builtin/staker/staker.go
+++ b/builtin/staker/staker.go
@@ -6,6 +6,7 @@
 package staker
 
 import (
+	"bytes"
 	"math/big"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


builtin/solidity/mapping.go
```diff
@@ -34,11 +34,13 @@ func (m *Mapping[K, V]) Get(key K) (value V, err error) {
 		if reflect.ValueOf(value).Kind() == reflect.Ptr {
 			value = reflect.New(reflect.TypeOf(value).Elem()).Interface().(V)
 		}
+		slots := (uint64(len(raw)) + 31) / 32
+		slots = max(1, slots)
+		m.context.UseGas(slots * thor.SloadGas)
+
```
otherview: Should we still charge if nothing was read ?

builtin/staker/staker.go
```diff
@@ -528,13 +521,46 @@ func (s *Staker) validateNextPeriodTVL(validator thor.Address) error {
...
-			logger.Debug("overrode state value", "variable", bytes32.String(), "value", o)
-			*ptr = o
-		}
+type configVar struct {
+	slot        thor.Bytes32
+	value       uint32
+	initialised bool
+}
+
+func newConfigVar(name string, defaultValue uint32) *configVar {
```
otherview: Good cleanup -I think this is a good candidate for the solidity package. wdyt ?

builtin/staker/staker.go
```diff
@@ -528,13 +521,46 @@ func (s *Staker) validateNextPeriodTVL(validator thor.Address) error {
...
+
+func (c *configVar) Name() string {
+	return string(bytes.TrimLeft(c.slot[:], string([]byte{0x00})))
+}
+
+func (c *configVar) Slot() thor.Bytes32 {
+	return c.slot
+}
+
+func debugOverride(sctx *solidity.Context, config *configVar) {
```
otherview: What if this was a method of the config variable struct ?

otherview (approved):
lgtm - Given it's refinement would consider moving it to solidity package, also helps to have less stuff in the staker.go, but none of this is a blocker imo

builtin/solidity/mapping.go
darrenkel: We were getting some `native_readMethod` that were using 0 gas. Felt a bit wrong to me?

otherview: Yeah I see what you mean, no strong feelings either way tbh - @libotony thoughts on this ?

libotony: In the EVM, SLOAD is using gas no matter the returned value is empty or not. So if we follow that, yes we should charge here.

libotony (approved):

YeahNotSewerSide (approved):

PR #1217, opened 2025-07-31
Move LinkedList to it's own service

otherview:
# Description

This PR refactors the staker linked list implementation to separate linked list operations from validation storage, improving performance and maintainability by eliminating redundant storage writes.

## Key Changes

- **New `linked_list` package**: Created dedicated package at `builtin/staker/linked_list/` for linked list operations
- **Separated storage concerns**: Linked list now manages its own Next/Prev pointer storage independently from validation data
- **Clean separation**: `Validation` struct no longer contains `Next`/`Prev` pointers

- **Eliminated redundant writes**: Removed the 3-write problem where linked list operations required updating multiple `Validation` records
- **Focused storage operations**: Linked list operations now only touch pointer storage, not validation data
- **Maintained API compatibility**: External interfaces remain unchanged

# Checklist:


builtin/staker/linkedlist/linked_list.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// NewLinkedList creates a new linked list with persistent storage mappings for staker management
+func NewLinkedList(sctx *solidity.Context, headPos, tailPos, countPos thor.Bytes32) *LinkedList {
+	return &LinkedList{
+		head:  solidity.NewAddress(sctx, headPos),
+		tail:  solidity.NewAddress(sctx, tailPos),
+		count: solidity.NewUint256(sctx, countPos),
+		next:  solidity.NewMapping[thor.Address, thor.Address](sctx, headPos),
+		prev:  solidity.NewMapping[thor.Address, thor.Address](sctx, tailPos),
```
libotony, line 32: The previous solution is straightforward also industrial standard and this one seems to be more complicated to understand.

otherview: Yes, the previous solution was used before, and I don't believe it makes sense to change the authority.go. It fits well in that case. I do believe the complexity of the staker.go is higher, but it's addressing a different set of requirements.

I'll ping you off-line to understand what makes this more complicated - there are likely improvements we can make together!

The context of this PR is a concern. My concern here was to decouple these two types of operations, ensuring that linked list operations are no longer tightly coupled with validator operations.

The previous implementation of the linked list in the staker package had a hard coupling between the validator object and the management of position for the linked list.
These 2 types of operations conflicted when doing their operations, and there were several cases where a validator was stored twice, `ActivateValidator` was one of them.

In general, a validator storing a change would require a SetValidation, but a linkedlist change also required a SetValidation.

I'm happy to take a different approach, I just want to address that concern.

libotony: Resolved offline.

builtin/staker/linkedlist/linked_list.go
```diff
@@ -0,0 +1,189 @@
...
+// Next returns the successor address in the list, or zero address if at the end
+func (l *LinkedList) Next(address thor.Address) (thor.Address, error) {
+	return l.next.Get(address)
+}
+
+// isHead checks if the given address is the head of the list
+func (l *LinkedList) isHead(address thor.Address) bool {
+	head, err := l.head.Get()
+	if err != nil {
+		return false
```
paologalligit: Small one: do we wanna still return the error to let the caller know what was wrong? I'm not sure but the caller might decide to do different things based on `false, nil` instead of `false, Error("...")`

otherview: Fair point, better safe than sorry 👍

paologalligit (approved):

PR #1221, opened 2025-08-01
Optimize solidity mapping perf

otherview:
# Description

This PR optimizes the generic `Mapping[K, V]` implementation by cutting down on allocations and CPU overhead in both `Get` and `Set`:

- **Pooled RLP buffers/readers** via `sync.Pool` to avoid per-call allocations of `bytes.Buffer` and `bytes.Reader`.
- **Fixed double-encode bug** in `Set` (now returns the pooled encoding instead of calling `rlp.EncodeToBytes` twice).
- **Nil-handling in `Get`** now only does a single reflection check on missing keys.

---

### Benchmark Comparison

| Case            | Old                                   | New                                   | Δ (ns/op)        | Δ (allocs)   | Δ (bytes)  |
| --------------- | ------------------------------------- | ------------------------------------- | ---------------- | ------------ | ---------- |
| **Set**         | 677 ns/op · 16 allocs · 475 B/op      | 550 ns/op · 11 allocs · 344 B/op      | –127 ns (~19%)   | –5           | –131 B      |
| **GetEmpty**    | 322 ns/op · 5 allocs · 168 B/op       | 313 ns/op · 5 allocs · 168 B/op       | –9 ns            | 0            | 0           |
| **GetNonEmpty** | 540 ns/op · 11 allocs · 416 B/op      | 535 ns/op · 10 allocs · 368 B/op      | –5 ns            | –1           | –48 B       |

---

### Benchmark test

```go
// mapping_bench_test.go
package solidity

import (
	"github.com/vechain/thor/v2/muxdb"
	"github.com/vechain/thor/v2/state"
	"github.com/vechain/thor/v2/trie"
	"math/big"
	"testing"

	"github.com/vechain/thor/v2/thor"
)

// stringKey implements the Key interface
type stringKey string

func (k stringKey) Bytes() []byte { return []byte(k) }

// ComplexVal is a sample value type containing *big.Int, uint32, and thor.Address
type ComplexVal struct {
	N *big.Int
	U uint32
	A thor.Address
}

func BenchmarkMapping(b *testing.B) {
	// 1) spin up an in-memory state.State (so you don’t hit disk or a real trie DB)
	//    adjust these calls to whatever the real constructor is
	st := state.New(muxdb.NewMem(), trie.Root{})

	// 2) you can pass `nil` for the charger if you don’t care about gas accounting
	ctx := NewContext(thor.Address{}, st, nil)

	m := NewMapping[stringKey, ComplexVal](ctx, thor.Bytes32{})

	key := stringKey("bench")
	missingKey := stringKey("missing-key")
	val := ComplexVal{}

	b.Run("Set", func(b *testing.B) {
		for i := 0; i < b.N; i++ {
			if err := m.Set(key, val, true); err != nil {
				b.Fatal(err)
			}
		}
	})

	b.Run("GetEmpty", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			if _, err := m.Get(missingKey); err != nil {
				b.Fatal(err)
			}
		}
	})

	// preload
	if err := m.Set(key, val, true); err != nil {
		b.Fatal(err)
	}

	b.Run("Get", func(b *testing.B) {
		b.ResetTimer()
		for i := 0; i < b.N; i++ {
			if _, err := m.Get(key); err != nil {
				b.Fatal(err)
			}
		}
	})
}
```

# Checklist:

### DIFF
--- a/builtin/solidity/mapping.go
+++ b/builtin/solidity/mapping.go
@@ -52,23 +52,35 @@ func NewMapping[K Key, V comparable](context *Context, pos thor.Bytes32) *Mappin
 
 // Get retrieves the value for the given key, charging SloadGas per 32-byte word.
 // If no value is present, returns the zero-value of V.
-func (m *Mapping[K, V]) Get(key K) (value V, err error) {
+func (m *Mapping[K, V]) Get(key K) (V, error) {
 	// compute the storage slot from key + base position
 	position := thor.Blake2b(key.Bytes(), m.basePos.Bytes())
 
-	err = m.context.state.DecodeStorage(m.context.address, position, func(raw []byte) error {
-		if len(raw) == 0 {
-			return nil
-		}
+	// prepare a zero-value container for decoding
+	var value V
 
-		// charge gas per 32-byte word
-		slots := (uint64(len(raw)) + 31) / 32
-		m.context.UseGas(slots * thor.SloadGas)
+	// attempt to decode storage into value
+	err := m.context.state.DecodeStorage(
+		m.context.address, position, func(raw []byte) error {
+			if len(raw) == 0 {
+				// no data, leave value as zero
+				return nil
+			}
 
-		// decode RLP in-place
-		return rlp.DecodeBytes(raw, &value)
-	})
-	return
+			// charge gas per 32-byte word
+			slots := (uint64(len(raw)) + 31) / 32
+			m.context.UseGas(slots * thor.SloadGas)
+
+			// decode RLP in-place
+			return rlp.DecodeBytes(raw, &value)
+		})
+	if err != nil {
+		// return zero type value
+		var zero V
+		return zero, err
+	}
+
+	return value, nil
 }
 
 // Set stores the given value at key, charging Sstore gas per 32-byte word.
@@ -83,6 +95,7 @@ func (m *Mapping[K, V]) Set(key K, value V, newValue bool) error {
 		return nil
 	}
 
+	// encode and store
 	return m.context.state.EncodeStorage(m.context.address, position, func() ([]byte, error) {
 		// encode via rlp library's internal pooling
 		buf, err := rlp.EncodeToBytes(value)


builtin/solidity/mapping.go
```diff
@@ -29,34 +31,83 @@ func NewMapping[K Key, V any](context *Context, pos thor.Bytes32) *Mapping[K, V]
...
+	if err := rlp.Encode(buf, v); err != nil {
+		return nil, err
+	}
+	// copy to independent slice for storage safety
+	out := make([]byte, buf.Len())
+	copy(out, buf.Bytes())
+	return out, nil
+}
+
+var readerPool = sync.Pool{
```
libotony: Nice catch, `encodeBuf` is perfect, but the `reader buf` will bring minor improvements as the major allocation is in `raw byte[]` and reader is just a container for that.

otherview: Thanks! :) 
What do you think we can/should improve further ?

builtin/solidity/mapping.go
```diff
@@ -29,34 +31,83 @@ func NewMapping[K Key, V any](context *Context, pos thor.Bytes32) *Mapping[K, V]
...
 func (m *Mapping[K, V]) Get(key K) (value V, err error) {
+	// compute the storage slot from key + base position
 	position := thor.Blake2b(key.Bytes(), m.basePos.Bytes())
+
 	err = m.context.state.DecodeStorage(m.context.address, position, func(raw []byte) error {
-		if reflect.ValueOf(value).Kind() == reflect.Ptr {
-			value = reflect.New(reflect.TypeOf(value).Elem()).Interface().(V)
-		}
 		if len(raw) == 0 {
+			// on missing-key, allocate a fresh pointer if V is a pointer type
```
otherview:
```suggestion
			// on missing-key, allocate a fresh pointer if V is a pointer type
			// this allows the calling of type.IsEmpty if not found 
```

builtin/solidity/mapping.go
otherview: Resolved offline

builtin/solidity/mapping.go
```diff
@@ -19,44 +17,86 @@ type Key interface {
...
 
-func NewMapping[K Key, V any](context *Context, pos thor.Bytes32) *Mapping[K, V] {
+// NewMapping creates a new persistent mapping at the given storage position.
+func NewMapping[K Key, V comparable](context *Context, pos thor.Bytes32) *Mapping[K, V] {
 	return &Mapping[K, V]{context: context, basePos: pos}
 }
 
+// Get retrieves the value for the given key, charging SloadGas per 32-byte word.
+// If no value is present, returns the zero-value of V.
 func (m *Mapping[K, V]) Get(key K) (value V, err error) {
```
paologalligit: Do you think this function would look clearer without the named returning values?

otherview: Likely yes, let me give it a go 👍

otherview: Yeah, I think it looks better. Perhaps a bit more verbose, but cleaner to read/understand 👍

PR #1238, opened 2025-08-08
feat(staker): move validate rules out of native implementation

libotony:
# Description

- Move validate errors to solidity implementation
- Isolate fatal error (mostly db/fs error) from user input error
- Move pause function to solidity
 
**If there is a fatal(state/db) error occurred during the native method execution, it means something wrong with the local node's database or filesystem, the tx should not be packed into the block. Other nodes in the network will disagree with the stateroot that the local node has produced.**

# Checklist:


builtin/gen/staker.sol
```diff
@@ -64,56 +62,74 @@ contract Staker {
...
      */
-    function increaseStake(
-        address validator
-    ) public payable checkStake(msg.value) {
-        string memory error = StakerNative(address(this)).native_increaseStake(
-            msg.sender,
-            validator,
-            msg.value
+    function increaseStake(address validator) public payable stakerNotPaused checkStake(msg.value) {
+        (address endorsor, , , , Status status, , , , uint32 exitBlock, ) = StakerNative(address(this))
```
darrenkel, line 66: Does this read the validator from state 3 times?

- `.native_getValidation(validator);`
- `.native_validateStakeChange(validator, msg.value, 0),`
- `.native_increaseStake(validator, msg.value);`

libotony: Yes, my planned next step is to work on an improvement of gas useage.

builtin/staker/validation/service.go
```diff
@@ -168,20 +159,6 @@ func (s *Service) Add(
...
-		return errors.New("stake is out of range")
-	}
-	val, err := s.GetValidation(validator)
-	if err != nil {
-		return err
-	}
-	if !val.IsEmpty() {
-		return errors.New("validator already exists")
-	}
-	if period != s.lowStakingPeriod && period != s.mediumStakingPeriod && period != s.highStakingPeriod {
```
darrenkel, line 181: Can't see this check if staker.sol, does it exist somewhere else?

darrenkel: ah, found it

darrenkel:
Personally feel like would be better to keep the contract as a proxy to native implementation and keep all business logic there, to keep it centralised. Seems now there can be a mix of .sol and .go validation

builtin/gen/staker.sol
libotony: Or you concerned about the performance?

builtin/staker/validation/service.go
libotony: https://github.com/vechain/thor/pull/1238/files#diff-4903ac9c9e3de129c7052884b81e60d1b4e01365d3c2de055722e5e04578ac00R230-R232

libotony:
> Personally feel like would be better to keep the contract as a proxy to native implementation and keep all business logic there, to keep it centralised. Seems now there can be a mix of .sol and .go validation

I understand your concern, I think the validating logic is not too much complicated, putting it into solidity also brings a benefit that computation gas is also counted. The native implementation should be as atomic as possible, the implementation of native contract should as slim as possible. Actually, the main focus of this PR is to isolate fatal error from validation, it lies in the solidity is just the logic seems simple enough to me. 

> Seems now there can be a mix of .sol and .go validation

Either in .sol, or in `staker_native.go`, not going further. I tried my best but `AddValidation` had a complicated logic, can be in solidity but seems fine to me if it is in `staker_native.go`.

Edited: `AddValidation` is also possible to fully moved to solidity, I'll take another look.

PR #1250, opened 2025-08-12
Vanja/chore/increase test coverage

vanja-vechain:
# Description

Add unit tests for delta, globalstats, stakes and validation

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:

### DIFF
--- a/builtin/staker/validation/validation.go
+++ b/builtin/staker/validation/validation.go
@@ -122,6 +122,15 @@ func (v *Validation) CurrentIteration() uint32 {
 func (v *Validation) Renew() *delta.Renewal {
 	newLockedVET := big.NewInt(0)
 
+	if v.QueuedVET == nil {
+		v.QueuedVET = new(big.Int)
+	}
+	if v.PendingUnlockVET == nil {
+		v.PendingUnlockVET = new(big.Int)
+	}
+	if v.WithdrawableVET == nil {
+		v.WithdrawableVET = new(big.Int)
+	}
 	newLockedVET.Add(newLockedVET, v.QueuedVET)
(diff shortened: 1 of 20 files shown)


builtin/staker/validation/validation.go
```diff
@@ -122,6 +122,15 @@ func (v *Validation) CurrentIteration() uint32 {
 func (v *Validation) Renew() *delta.Renewal {
 	newLockedVET := big.NewInt(0)
 
+	if v.QueuedVET == nil {
```
paologalligit, line 125: Would it make sense to have a `New()` function where all these values are initialized? Or they might be null during their lifecycle?

paologalligit (approved):
Looks good, just left a question

builtin/staker/validation/validation.go
darrenkel: IMO if `if v.QueuedVET == nil {` is true then the state has been corrupted and should be more of a panic situation

darrenvechain (approved):

builtin/staker/validation/validation.go
paologalligit: yes, you're right, the `queuedVET` should never be 0 when calling `Renew()`. But as for `PendingUnlockVET` and `WithdrawableVET`, they could be 0 in a non-corrupted state, right?

vanja-vechain: Yes, this was just caught in unit test, I think in normal operation they are always populated, just added safety

PR #1251, opened 2025-08-13
Change stop energy condition

paologalligit:
# Description

VTHO generation must stop on the hardfork block, not on the PoS transition block. This PR updates the condition and add more tests.

[GithubIssue](https://github.com/vechain/protocol-board-repo/issues/755)

# Checklist:


vanja-vechain:
There are couple of places in codebase where actual transition is identified by evaluating energy stop growth key, those should be reviewed since now energy stop growth is not a point of transition

consensus/consensus.go
```diff
@@ -89,17 +89,16 @@ func (c *Consensus) NewRuntimeForReplay(header *block.Header, skipValidation boo
...
 			return nil, err
 		}
 		if len(activeGroup) > 0 {
 			// invalidate cache
 			c.validatorsCache.Add(header.ParentID(), activeGroup)
 		}
-		if activated {
-			err := builtin.Energy.Native(state, parentSummary.Header.Timestamp()).StopEnergyGrowth()
-			if err != nil {
+		if header.Number() == c.forkConfig.HAYABUSA {
```
darrenkel: @paologalligit maybe we could put this in the runtime package, where we set the hayabusa code

https://github.com/vechain/thor/blob/release/hayabusa/runtime/runtime.go#L141

darrenkel: Same for packer.go

darrenkel: FYI the reason for not putting the transition in there is because we need to decide if the block proposer is PoA or dPoS before we init the runtime

paologalligit: you mean move these into runtime or add it there too?

darrenkel: I mean move `StopEnergyGrowth()` to runtime. Instead of having it both consensus and packer packages. Its not conditional on pos anymore

vanja-vechain:
> There are couple of places in codebase where actual transition is identified by evaluating energy stop growth key, those should be reviewed since now energy stop growth is not a point of transition

Not a concern anymore

vanja-vechain (approved):

PR #1255, opened 2025-08-13
Vanja/feat/706 validator probability weight to 1 instead of 2 when no delegation applied

vanja-vechain:
# Description

Refactor validator multiplier, if there are no delegations validator weight multiplier is 1, 2 otherwise

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:


builtin/staker/aggregation/service.go
```diff
@@ -110,16 +110,20 @@ func (s *Service) Exit(validator thor.Address) (*delta.Exit, error) {
 
 // SignalExit marks locked delegations as exiting for the next period.
 // Called when a validator signals intent to exit but hasn't exited yet.
-func (s *Service) SignalExit(validator thor.Address, stake *stakes.WeightedStake) error {
+func (s *Service) SignalExit(validator thor.Address, stake *stakes.WeightedStake, valStake *big.Int) error {
```
darrenkel, line 113: Why do we add the valStake here? I think aggregation should represent only the sum of delegators.

darrenkel: Ah, think I understand it. The extra weight is assigned to the aggregation if there is a delegation?

Might be harder to manage stake increases / decreases?

darrenkel: IMO, I think its better to only put this in the weight field, because we can always tell if there are delegations when setting the weight

darrenkel: Not so trivial, had a look at the codebase

darrenkel: Any solution is gonna make global stats very messed up

builtin/staker/validations_test.go
```diff
@@ -2530,3 +2534,397 @@ func getTestMaxLeaderSize(param *params.Params) *big.Int {
...
+	assert.Equal(t, big.NewInt(0).Add(stake, big.NewInt(0).Mul(dStake.VET(), big.NewInt(2))), totals2.TotalQueuedStake)
+	assert.Equal(t, big.NewInt(0).Add(
+		big.NewInt(0).Mul(stake, big.NewInt(2)),
+		big.NewInt(0).Mul(dStake.Weight(), big.NewInt(2))),
+		totals2.TotalQueuedWeight,
+	)
+
+	// exit queued
+	staker.SignalDelegationExit(delegationID3)
+	staker.WithdrawDelegation(delegationID3)
```
darrenkel: Does this not error without running housekeeping in between?

darrenkel: Looks like withdraw does fail actually

darrenkel:
```
newTestSequence(t, staker).SignalDelegationExit(delegationID3).WithdrawDelegation(delegationID3, nil)
```

vanja-vechain: Fixed now

darrenvechain (approved):

PR #1268, opened 2025-08-19
add system level config for change some const parameters

libotony:
# Description

This PR makes some constant parameter to be configurable either during test or through custom genesis. Which will make test easier to run and also reduced some storage slot accessing in staker contract, which will be an overhead for main/test net.

**Since config is in memory, make sure it is locked ASAP for any production network.**

# Checklist:


thor/config.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	if cfg.CooldownPeriod != 0 {
+		CooldownPeriod = cfg.CooldownPeriod
+	}
+}
+
+// LockConfig locks the config, preventing any further changes.
+// Required for mainnet and testnet.
+func LockConfig() {
```
darrenkel, line 82: This doesn't really prevent since client could do:

```go
thor.BlockInterval = 5
```


Should we do 
```
var (
	blockInterval              uint64 = 10   
)
func BlockInterval() uint64 { return  blockInterval }
```

vanja-vechain (approved):

thor/config.go
libotony: You're right, this is better, will append a commit

PR #1280, opened 2025-08-21
Sched proposer check

otherview:
# Description

Check Proposer Weights instead of weight float normalization

Fixes # [(issue)](https://github.com/vechain/otherviews-workboard/issues/206)

# Checklist:


pos/sched.go
```diff
@@ -56,13 +58,28 @@ func NewScheduler(
...
+
 	online := make([]*onlineProposer, 0)
 	for id, p := range proposers {
 		if id == addr {
 			proposer = p
 			proposerID = id
 		}
 		if p.OfflineBlock == nil || id == addr {
+			// ensure staker must have a minimum stake
+			if p.Weight.Cmp(minWeight) < 0 {
```
darrenkel: Whats the benefit here? seems like we'd just freeze block production and this requirement should only exist within staker package IMO

darrenkel: Maybe don't return the error, just don't add the validator to the sequence

otherview: resolved offline 👍

PR #1281, opened 2025-08-22
move double signing detaction to post block process hook

libotony:
# Description

# Checklist:

### DIFF
--- a/cmd/thor/node/node.go
+++ b/cmd/thor/node/node.go
@@ -282,13 +282,14 @@ func (n *Node) txStashLoop(ctx context.Context) {
 }
 
 // guardBlockProcessing adds lock on block processing and maintains block conflicts.
-func (n *Node) guardBlockProcessing(blockNum uint32, process func(conflicts [][]byte) (thor.Bytes32, error)) error {
+func (n *Node) guardBlockProcessing(blockNum uint32, process func(conflicts uint32) (thor.Bytes32, error)) error {
 	n.processLock.Lock()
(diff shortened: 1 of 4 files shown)


vanja-vechain (approved):

cmd/thor/node/node.go
```diff
@@ -336,45 +337,52 @@ func (n *Node) guardBlockProcessing(blockNum uint32, process func(conflicts [][]
...
 				// logic to verify that the blocks are different and from the same signer
-				signer, err := conflictBlock.Header().Signer()
+				existingSigner, err := conflictBlock.Header().Signer()
 				if err != nil {
-					return thor.Bytes32{}, err
+					return err
 				}
-				if signer == newSigner &&
-					conflictBlock.Header().ID() != newBlock.Header().ID() &&
-					conflictBlock.Header().StateRoot() != newBlock.Header().StateRoot() {
```
YeahNotSewerSide, line 371: Hey, @libotony should we leave a check for the state roots? Initial idea was that if the transactions are the same in 2 blocks(they converge to the same state) then it's not really a double signing, even if IDs are different for some reason, since no MEV being extracted

libotony: One scenario I was thinking is they can also manipulate the BFT voting with state root kept the same.

YeahNotSewerSide: I see

YeahNotSewerSide (approved):

PR #1303, opened 2025-09-01
update gas usage of raw

libotony:
# Description

# Checklist:


builtin/staker/globalstats/service.go
```diff
@@ -57,11 +58,17 @@ func (s *Service) ApplyRenewal(renewal *Renewal) error {
 	locked.Sub(renewal.LockedDecrease)
 	queued -= renewal.QueuedDecrease
 
-	if err := s.locked.Set(locked); err != nil {
-		return err
+	if empty {
```
vanja-vechain: Why not upsert here?

libotony: Yes, we can do `upsert` here, just trying to balance from style and performance. `Upsert` will have an extra read(cached in most scenario, so an extra rlp decode), probably this is fine, I might be overthinking. BTW, I do like using `upsert`.

PR #1307, opened 2025-09-01
add toWordSize

libotony:
# Description

Previously, we are charging slot accessing and storing by word size, which is per-32 byte, this is what solidity compiler is doing, but the compiler are splitting big types into 32-byte storage slots, which result in multiple disk read.

But in our implementation, it's a bit unfair to user if we charge per 32-byte since every accessing is one disk read. So this PR proposes maximum 2 word cost for each slot. And this also relfected in the `staker_native_gas_test.go`.

# Checklist:

### DIFF
--- a/builtin/solidity/mapping.go
+++ b/builtin/solidity/mapping.go
@@ -23,23 +23,20 @@ type Key interface {
 //   - Setting the default type value means removing the value from storage
 //
 //   - If Mapping[thor.Bytes32, *thor.Address]:
-//     ex: SET (k:0x123, v:nil) -> stores nil
-//     ex: SET (k:0x123, v:&thor.Address{})  -> stores nil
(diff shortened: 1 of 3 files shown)


vanja-vechain (approved):

builtin/solidity/mapping.go
```diff
@@ -23,23 +23,20 @@ type Key interface {
 //   - Setting the default type value means removing the value from storage
 //
 //   - If Mapping[thor.Bytes32, *thor.Address]:
-//     ex: SET (k:0x123, v:nil) -> stores nil
-//     ex: SET (k:0x123, v:&thor.Address{})  -> stores nil
```
otherview, line 27: I wouldn't delete this as it makes clear that the default instance value of a pointer is stored as nil.

libotony: Resolved offline.

PR #1327, opened 2025-09-05
chore: `GetMaxBlockProposers` extracted

freemanzMrojo:
# Description

Extracted duplicated logic into a single method.

Difference with PoA taken into account for cases where `maxBlockProposers > thor.InitialMaxBlockProposers` (`capToInitial` = `true`)

### DIFF
--- a/thor/params.go
+++ b/thor/params.go
@@ -70,7 +70,8 @@ var (
 // If capToInitial is true, values greater than InitialMaxBlockProposers are also capped (PoA)
 func GetMaxBlockProposers(params interface {
 	Get(Bytes32) (*big.Int, error)
-}, capToInitial bool) (uint64, error) {
+}, capToInitial bool,
+) (uint64, error) {
 	mbp, err := params.Get(KeyMaxBlockProposers)
 	if err != nil {
 		return 0, err
--- a/thor/params.go
+++ b/thor/params.go
@@ -65,3 +65,20 @@ var (
 	EnergyGrowthRate      = big.NewInt(5000000000) // WEI THOR per token(VET) per second. about 0.000432 THOR per token per day.
 	NumberOfBlocksPerYear = big.NewInt(8640 * 365) // number of blocks per year, non leap (365 days)
 )
+
+// GetMaxBlockProposers retrieves the max block proposers parameter with fallback to initial value
+// If capToInitial is true, values greater than InitialMaxBlockProposers are also capped (PoA)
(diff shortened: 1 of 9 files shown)


thor/params.go
```diff
@@ -65,3 +65,21 @@ var (
 	EnergyGrowthRate      = big.NewInt(5000000000) // WEI THOR per token(VET) per second. about 0.000432 THOR per token per day.
 	NumberOfBlocksPerYear = big.NewInt(8640 * 365) // number of blocks per year, non leap (365 days)
 )
+
+// GetMaxBlockProposers retrieves the max block proposers parameter with fallback to initial value
+// If capToInitial is true, values greater than InitialMaxBlockProposers are also capped (PoA)
+func GetMaxBlockProposers(params interface {
```
paologalligit, line 71: Do you think this function could receive the default max block proposers value instead of the first interface? This way it only holds the following logic that could change the value based on the `capToInitial`, which is the duplicated one. Also, if the `cappedToInitial` is something bounded to `PoA`, we could rename it something more explicit, like `isPoA`

paologalligit (commented):
Now that we have extracted the method we can also add a unit test for it that tests the different conditions

thor/params.go
freemanzMrojo: I think it is better the existing way. Otherwise we would need to repeat this block:

```
    // Where params can be different things
	mbp, err := params.Get(KeyMaxBlockProposers)
	if err != nil {
		return 0, err
	}
	
	// And now thor.GetMaxProposers(mbp)
```

With the current approach is a one-liner.

freemanzMrojo: `capToInitial` looks indeed related to PoA but with the current name is more generic. Otherwise is bound to PoA. Not a hard requirement on my side though, I can change the name if needs be.

darrenkel, line 71: Should we just implement this function in params.go ? eg. 
```builtin.Params.Native(state).GetMaxBlockProposers(...)```

freemanzMrojo: I thought of that but right now the list of methods of that `params.go` is pretty generic (just `Get` and `Set`) and this `MaxBlockProposers` is pretty concrete.

I do not mind in the end, but since it applies to 2 parameters at `thor/params.go` I think the current place is a better home.

freemanzMrojo:
> Now that we have extracted the method we can also add a unit test for it that tests the different conditions

Sure, I will add them after the other stuff gets clarified/agreed.

PR #1331, opened 2025-09-09
chore: `GetMaxBlockProposers` extracted

freemanzMrojo:
# Description

Extracted duplicated logic into a single method.

Difference with PoA taken into account for cases where `maxBlockProposers > thor.InitialMaxBlockProposers` (`capToInitial` = `true`)

### DIFF
--- a/builtin/staker/housekeep.go
+++ b/builtin/staker/housekeep.go
@@ -97,6 +97,19 @@ func (s *Staker) evictionCallback(currentBlock uint32, evictions *[]thor.Address
 	}
 }
 
+// getMaxBlockProposers returns the max block proposers from the params
+func (s *Staker) getMaxBlockProposers() (uint64, error) {
+	mbp, err := s.params.Get(thor.KeyMaxBlockProposers)
+	if err != nil {
+		return 0, err
+	}
+	maxBlockProposers := mbp.Uint64()
+	if maxBlockProposers == 0 {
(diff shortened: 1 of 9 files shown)


builtin/staker/housekeep.go
```diff
@@ -113,14 +113,10 @@ func (s *Staker) computeActivationCount(hasValidatorExited bool) (uint64, error)
 		leaderSize = leaderSize - 1
 	}
 
-	mbp, err := s.params.Get(thor.KeyMaxBlockProposers)
+	maxBlockProposers, err := thor.GetMaxBlockProposers(s.params, false)
 	if err != nil {
 		return 0, err
 	}
-	maxBlockProposers := mbp.Uint64()
```
vanja-vechain, line 120: I think this ones are needed for existing networks, since it is now in config. And for testnet and mainnet config is set to be locked

freemanzMrojo: Not sure if I follow what you mean. I havent changed the functionality, just extracted the repeated code to one place. So the result of `thor.GetMaxBlockProposers` should return the same value as the previous version.

vanja-vechain (approved):

builtin/staker/housekeep.go
vanja-vechain: Makes sense, just saw function there

PR #1358, opened 2025-09-18
improve globalstats

libotony:
# Description
 
+ Tidy up `globalstatus` 
+ Add overflow check in `WeightedStake`
+ Add `Validation.CooldownEnded`
+ Fixed a bug in `Validation.renew()`, which is PendingUnlock and Queued is not correctly reflected in `renewal`.

# Checklist:

### DIFF
--- a/builtin/staker/aggregation/aggregation.go
+++ b/builtin/staker/aggregation/aggregation.go
@@ -59,18 +59,15 @@ func (a *Aggregation) renew() (*globalstats.Renewal, error) {
 	queuedDecrease := a.Pending.VET
 
 	// Move Pending => Locked
-	err := a.Locked.Add(a.Pending)
-	if err != nil {
+	if err := a.Locked.Add(a.Pending); err != nil {
 		return nil, err
 	}
 	a.Pending = &stakes.WeightedStake{}
 
 	// Remove ExitingVET from LockedVET
-	err = a.Locked.Sub(a.Exiting)
-	if err != nil {
+	if err := a.Locked.Sub(a.Exiting); err != nil {
 		return nil, err
 	}
-
 	a.Exiting = &stakes.WeightedStake{}
 
 	return &globalstats.Renewal{
--- a/builtin/staker/validation/validation.go
+++ b/builtin/staker/validation/validation.go
@@ -186,11 +186,8 @@ func (v *Validation) renew(delegationWeight uint64) (*globalstats.Renewal, error
 	}
 	prev.valWeight = stakes.NewWeightedStakeWithMultiplier(v.LockedVET, v.multiplier()).Weight
 
-	lockedIncrease := stakes.NewWeightedStake(0, 0)
-	lockedDecrease := stakes.NewWeightedStake(0, 0)
-
-	lockedIncrease.VET = v.QueuedVET
-	lockedDecrease.VET = v.PendingUnlockVET
+	lockedIncrease := stakes.NewWeightedStake(v.QueuedVET, 0)
+	lockedDecrease := stakes.NewWeightedStake(v.PendingUnlockVET, 0)
 
 	v.LockedVET += v.QueuedVET
 	var underflow bool
@@ -251,10 +248,7 @@ func (v *Validation) exit() *globalstats.Exit {
 
 // CooldownEnded returns true if validator has exited and the cooldown period has ended.
 func (v *Validation) CooldownEnded(currentBlock uint32) bool {
-	if v.ExitBlock != nil && *v.ExitBlock+thor.CooldownPeriod() <= currentBlock {
-		return true
-	}
-	return false
+	return v.ExitBlock != nil && *v.ExitBlock+thor.CooldownPeriod() <= currentBlock
 }
 
 // CalculateWithdrawableVET returns the validator withdrawable amount for a given block + period


builtin/staker/aggregation/aggregation.go
```diff
@@ -59,17 +59,18 @@ func (a *Aggregation) renew() (*globalstats.Renewal, error) {
...
+		return nil, err
+	}
 	a.Pending = &stakes.WeightedStake{}
 
 	// Remove ExitingVET from LockedVET
-	err := a.Locked.Sub(a.Exiting)
+	err = a.Locked.Sub(a.Exiting)
 	if err != nil {
 		return nil, err
 	}
```
otherview: nit:
```suggestion
	if err := a.Locked.Sub(a.Exiting); err != nil {
		return nil, err
	}
```

builtin/staker/aggregation/aggregation.go
```diff
@@ -59,17 +59,18 @@ func (a *Aggregation) renew() (*globalstats.Renewal, error) {
 	queuedDecrease := a.Pending.VET
 
 	// Move Pending => Locked
-	a.Locked.Add(a.Pending)
+	err := a.Locked.Add(a.Pending)
+	if err != nil {
+		return nil, err
+	}
```
otherview: nit:
```suggestion
	if err := a.Locked.Add(a.Pending); err != nil {
		return nil, err
	}
```

builtin/staker/globalstats/service.go
```diff
@@ -58,68 +58,78 @@ func (s *Service) ApplyRenewal(renewal *Renewal) error {
...
+	if aggregation.ExitedTVL.VET > 0 {
+		if err := s.AddWithdrawable(aggregation.ExitedTVL.VET); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
+// RemovedLocked decreases locked totals when stake is removed from the locked.
```
otherview: nit
```suggestion
// RemoveLocked decreases locked totals when stake is removed from the locked.
```

builtin/staker/staker.go
```diff
@@ -403,29 +403,31 @@ func (s *Staker) WithdrawStake(validator thor.Address, endorser thor.Address, cu
...
+	}
+	if cooldownVET > 0 {
+		if err = s.globalStatsService.RemoveCooldown(cooldownVET); err != nil {
 			return 0, err
 		}
 	}
 
 	logger.Info("withdrew validator staker", "validator", validator)
-	return stake, nil
+	return withdrawableVET + queuedVET + cooldownVET, nil
```
freemanzMrojo, line 434: Out of curiosity, why this is the result instead of just `withdrawableVET`?

builtin/staker/validation/validation.go
```diff
@@ -254,12 +249,19 @@ func (v *Validation) exit() *globalstats.Exit {
 	}
 }
 
+// CooldownEnded returns true if validator has exited and the cooldown period has ended.
+func (v *Validation) CooldownEnded(currentBlock uint32) bool {
+	if v.ExitBlock != nil && *v.ExitBlock+thor.CooldownPeriod() <= currentBlock {
```
freemanzMrojo: Minor: Maybe we could just return the condition instead of `if - else`

builtin/staker/validation/validation.go
```diff
@@ -180,43 +181,37 @@ func (v *Validation) renew(delegationWeight uint64) (*globalstats.Renewal, error
...
 		multiplier uint8
 	}
-	prev.lockedVET = v.LockedVET
 	prev.valWeight = stakes.NewWeightedStakeWithMultiplier(v.LockedVET, v.multiplier()).Weight
 
+	lockedIncrease := stakes.NewWeightedStake(0, 0)
+	lockedDecrease := stakes.NewWeightedStake(0, 0)
+
+	lockedIncrease.VET = v.QueuedVET
+	lockedDecrease.VET = v.PendingUnlockVET
```
freemanzMrojo: Couple of comments:

1. Maybe we can add these assignments directly in the constructor `NewWeightedStake`
2. Why the previous logic is no longer being applied?

```
if prev.lockedVET < after.lockedVET {
		lockedIncrease.VET = after.lockedVET - prev.lockedVET
	} else {
		lockedDecrease.VET = prev.lockedVET - after.lockedVET
	}
```

otherview (approved):

builtin/staker/staker.go
libotony: We need to know exactly each of these value `withdrawable/queuedVET/cooldown` to update into `globalstats`, so returned them all and do the sum here.

builtin/staker/validation/validation.go
libotony: This is a messy logic, previously brought by the multiplier change from 2->1 or 1->2(with or without delegation). The locked increase VET has it's own pattern can follow, which is locked increase is pending stake and locked decrease is pending unlock, and subsequent calls relying on locked decrease to reflect to pending unlock, so changed it back.

And good find on point 1, will append a commit.

paologalligit (approved):

builtin/staker/staker.go
freemanzMrojo: Thanks @libotony !

builtin/staker/validation/validation.go
freemanzMrojo: Perfect, thanks!

PR #1362, opened 2025-09-18
Paolo/sanity check/814 bis

paologalligit:
# Description

Ensure that locked + queued + withdrawable VET + cooldown VET equals the total VET in the staker account address when calling Housekeep and apply transactions.

[GithubIssue](https://github.com/vechain/protocol-board-repo/issues/814)

# Checklist:

### DIFF
--- a/builtin/solidity/context.go
+++ b/builtin/solidity/context.go
@@ -34,3 +34,7 @@ func (c *Context) UseGas(gas uint64) {
 		c.charger.Charge(gas)
 	}
 }
+
+func (c *Context) Address() thor.Address {
+	return c.address
+}
--- a/builtin/staker/housekeep.go
+++ b/builtin/staker/housekeep.go
@@ -172,7 +172,7 @@ func (s *Staker) applyEpochTransition(transition *EpochTransition) error {
 		logger.Info("exiting validator", "validator", transition.ExitValidator)
 
 		// Now call ExitValidator to get the actual exit details and perform the exit
-		exit, err := s.validationService.ExitValidator(transition.ExitValidator)
+		valExit, err := s.validationService.ExitValidator(transition.ExitValidator)
 		if err != nil {
 			return err
 		}
@@ -182,8 +182,7 @@ func (s *Staker) applyEpochTransition(transition *EpochTransition) error {
 			return err
 		}
 
-		valExitedTVL := exit.ExitedTVL.VET
-		if err := s.globalStatsService.ApplyExit(valExitedTVL, exit.Add(aggExit)); err != nil {
+		if err := s.globalStatsService.ApplyExit(valExit, aggExit); err != nil {
 			return err
 		}
 	}
@@ -236,8 +235,13 @@ func (s *Staker) activateNextValidation(currentBlk uint32, maxLeaderGroupSize ui
 		return thor.Address{}, err
 	}
 
+	globalRenewal, err := validatorRenewal.Add(aggRenew)
+	if err != nil {
+		return thor.Address{}, err
+	}
+
 	// Update global stats with both validator and delegation renewals
-	if err = s.globalStatsService.ApplyRenewal(validatorRenewal.Add(aggRenew)); err != nil {
+	if err = s.globalStatsService.ApplyRenewal(globalRenewal); err != nil {
 		return thor.Address{}, err
 	}
 
--- a/builtin/staker/housekeep.go
+++ b/builtin/staker/housekeep.go
@@ -6,6 +6,8 @@
 package staker
 
 import (
+	"fmt"
+
 	"github.com/vechain/thor/v2/builtin/staker/globalstats"
 	"github.com/vechain/thor/v2/builtin/staker/validation"
 	"github.com/vechain/thor/v2/thor"
@@ -48,6 +50,11 @@ func (s *Staker) Housekeep(currentBlock uint32) (bool, error) {
 	if err := s.applyEpochTransition(transition); err != nil {
 		return false, err
 	}
+
+	if err := s.PerformSanityCheck(); err != nil {
+		return false, err
+	}
+
 	return true, nil
 }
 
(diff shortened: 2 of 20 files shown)


builtin/staker/housekeep.go
```diff
@@ -247,3 +263,45 @@ func (s *Staker) activateNextValidation(currentBlk uint32, maxLeaderGroupSize ui
...
+	}
+
+	// Get the staker contract's account balance
+	stakerAddr := s.validationService.ContractAddress()
+	balance, err := s.state.GetBalance(stakerAddr)
+	if err != nil {
+		return err
+	}
+	balanceVET := ToVET(balance)
+	total := lockedStake + queuedStake + withdrawableStake + cooldownStake
```
vanja-vechain: I guess this could be flagged as potential overflow point, we know all of those are pre-checked but not the sum

builtin/staker/staker.go
```diff
@@ -414,13 +414,13 @@ func (s *Staker) WithdrawStake(validator thor.Address, endorser thor.Address, cu
 	}
 
 	// update global stats
-	if withdrawableVET > 0 {
-		if err = s.globalStatsService.RemoveWithdrawable(withdrawableVET); err != nil {
+	if withdrawableVET > queuedVET+cooldownVET {
```
vanja-vechain: This will be resolved by Darren's PR then

builtin/staker/validation/service.go
```diff
@@ -84,6 +84,17 @@ func (s *Service) LeaderGroupIterator(callbacks ...func(thor.Address, *Validatio
 	})
 }
 
+func (s *Service) QueuedGroupIterator(callbacks ...func(thor.Address, *Validation) error) error {
```
vanja-vechain: Seems like this is not used

builtin/staker/validation/repository.go
```diff
@@ -162,6 +162,10 @@ func (r *Repository) iterateActive(callback func(thor.Address, *Validation) erro
 	return r.activeList.Iterate(callback)
 }
 
+func (r *Repository) iterateQueued(callback func(thor.Address, *Validation) error) error {
```
vanja-vechain: This one as well

builtin/staker/validation/service.go
```diff
@@ -469,3 +480,7 @@ func makeRewardKey(validator thor.Address, stakingPeriod uint32) thor.Bytes32 {
 	copy(key[4:], validator.Bytes())
 	return key
 }
+
+func (s *Service) ContractAddress() thor.Address {
```
libotony: This is not necessary, the address is passed from the constructor `Staker.New(address`. Normally, it's `builtin.Staker.Address`.

consensus/validator.go
```diff
@@ -303,6 +303,9 @@ func (c *Consensus) verifyBlock(blk *block.Block, state *state.State, blockConfl
 
 	if posActive {
 		staker := builtin.Staker.Native(state)
+		if err := staker.PerformSanityCheck(); err != nil {
```
libotony: Wrap the error by `consensusError` will be better.

builtin/staker_native_gas_test.go
```diff
@@ -135,7 +135,7 @@ func TestStakerNativeGasCosts(t *testing.T) {
 		},
 		{
 			function:     "native_withdrawStake",
-			expectedGas:  16200,
+			expectedGas:  11200,
```
libotony, line 138: After this change, test will fail?

builtin/staker/validation/service.go
libotony: After looking at the code where it was used, maybe adding address or sctx to the staker struct is better.

consensus/validator.go
darrenkel: Maybe we can just leave this in the housekeeping, where we can revert the state if the check is not sane

darrenkel:
```go
	dPosStatus, err := staker.SyncPOS(c.forkConfig, header.Number())
	if err != nil {
		log.Error("staker sync pos failed - reverting state", "err", err, "height", header.Number(), "parent", parent, "checkpoint", checkpoint)
		dPosStatus.Updates = false // reset since no changes actually occurred
		state.RevertTo(checkpoint)
	}
```

SyncPOS will perform the sanity check via housekeeping

builtin/solidity/context.go
```diff
@@ -34,3 +34,7 @@ func (c *Context) UseGas(gas uint64) {
 		c.charger.Charge(gas)
 	}
 }
+
+func (c *Context) Address() thor.Address {
+	return c.address
+}
```
otherview: Probably ok to store the address or the `sctx := solidity.NewContext(addr, state, charger)` in a property of the Staker Object

builtin/staker/housekeep.go
paologalligit: good point, adding safe sum

builtin/staker/validation/service.go
paologalligit: yep, it was something I added at first and forgot to remove 👍

builtin/staker_native_gas_test.go
paologalligit: hmm, what do you mean? the test was failing because of the old value `16200`

builtin/staker/validation/service.go
paologalligit: I'm using the address received as input in the `Staker` constructor. Let me know if you're fine with this solution

consensus/validator.go
paologalligit: I'm fine with either solutions: it's a consensus error, so I'm ok if we wrap it like that. Then we have the description that's telling us that the balance sanity check failed

builtin/solidity/context.go
otherview: ok to remove, no longer used it seems.

builtin/staker/housekeep.go
```diff
@@ -6,11 +6,24 @@
...
 )
 
+func ToVET(wei *big.Int) uint64 {
+	return new(big.Int).Div(wei, bigE18).Uint64()
+}
+
+func ToWei(vet uint64) *big.Int {
+	return new(big.Int).Mul(new(big.Int).SetUint64(vet), bigE18)
+}
+
```
otherview: Might move it to the `protocol.go` file, most funcs there are only called by the protocol.

builtin/staker/housekeep.go
```diff
@@ -247,3 +265,58 @@ func (s *Staker) activateNextValidation(currentBlk uint32, maxLeaderGroupSize ui
 
 	return validator, nil
 }
+
+// This check ensures that locked + queued + withdrawable VET equals the total VET in the staker account address.
```
otherview:
```suggestion
// PerformSanityCheck checks ensures that locked + queued + withdrawable VET equals the total VET in the staker account address.
```

builtin/staker/housekeep.go
```diff
@@ -247,3 +265,58 @@ func (s *Staker) activateNextValidation(currentBlk uint32, maxLeaderGroupSize ui
...
+func (s *Staker) PerformSanityCheck() error {
```
otherview: nit: Can this name be changed to something like `BalanceCheck` or `ContractBalanceCheck` or something a bit more descriptive ?

vanja-vechain (approved):

builtin/staker/housekeep.go
paologalligit: yeah, makes more sense

vanja-vechain (approved):

PR #1364, opened 2025-09-18
change queuedStake to stakes in staker sol

libotony:
# Description

To reflect our recent change, returns locked, queued, withdrawable, cooldown in one function

# Checklist:


freemanzMrojo:
Just a general comment @libotony , as far as I see with the change there is no place in the code - for now - that reads something different to the queued stake. Is this right?

IMO if this is merged another ticket should be created to refactor `native_totalStake` since it returns LockedVET (that is also returned by the new `native_stakes`) and the LockedWeight.

To sum up, is this PR necessary for today or it can wait to be merged?

Thanks in advance.

builtin/gen/staker.sol
```diff
@@ -34,11 +34,11 @@ contract Staker {
     }
 
     /**
-     * @dev queuedStake returns all stakes by queued validators.
+     * @dev stakes returns the amount of VET locked, queued, withdrawable and cooldown of all validations and delegations.
      */
-    function queuedStake() public view returns (uint256 queuedVET) {
-        return StakerNative(address(this)).native_queuedStake();
-    }
+    function stakes() public view returns (uint256 locked, uint256 queued, uint256 withdrawable, uint256 cooldown) {
```
darrenkel, line 39: IMO we should not return cooldown and withdrawable. Those values may not be accurate.

darrenkel: The reason being, once a validators cooldown period is over, their funds are technically withdrawable, so cooldown and withdrawable counters are out of sync. Would be better to leave them internal I think.

We don't have a process to move them over

vanja-vechain: Agree with @darrenvechain here, there could be periods when figures can be out of sync, the cooldown resolution is actually happening only when they withdraw

otherview:
Looks like what is currently available to `Withdraw` from the contract is not strictly tied to `withdrawable` counter.

Adding this method would create a caveat where the numbers in `withdrawable` and `cooldown` while representing the exact numbers, don't represent the contract state. Some of the `cooldown` might be actually `withdrawable`.

There's a concern here where this would provide information on the counters and be read as information on the contract state.

For data handling, these figures by directly looking at the storage, and there is no direct requirement of supplying this information. It's probably ok to defer this to another time. 

I understand that changes to the staker contract will mean a hardfork, but anyone looking to retrieve the contract current state should be aware of the `withdrawable` + `cooldown` caveats.

libotony:
> Just a general comment @libotony , as far as I see with the change there is no place in the code - for now - that reads something different to the queued stake. Is this right?
> 
> IMO if this is merged another ticket should be created to refactor `native_totalStake` since it returns LockedVET (that is also returned by the new `native_stakes`) and the LockedWeight.
> 
> To sum up, is this PR necessary for today or it can wait to be merged?
> 
> Thanks in advance.

No rush on that, looks we have more concerns than the features it brings, probably not going to merge it.

PR #1372, opened 2025-09-23
chore(`modernize`): updated and fixed check

freemanzMrojo:
# Description

Lately It was not detecting correctly the improvement suggestions.

Job checking this out [here](https://github.com/vechain/thor/actions/runs/17938753485/job/51010070835?pr=1372).

### DIFF
--- a/muxdb/trie_test.go
+++ b/muxdb/trie_test.go
@@ -70,7 +70,7 @@ func TestTrieVersioning(t *testing.T) {
 		roots []trie.Root
 	)
 
-	for i := uint32(0); i < 5; i++ {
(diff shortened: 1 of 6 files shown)


freemanzMrojo:
In principle it should be ok, the tool detects the go version beforehand so 1.24 for now on our side. I thought the same about using a static version but I got convinced.

> Looks good to me, just a thought: if we are relying on latest, does it mean that we might get linter errors from a newer go version we're not using? (e.g. the latest modernize tool starts to address code features in go 1.25 while we still are 1.24 and we don't intend to upgrade in short time). Maybe it's better to set a more static version based on the internal go version?

PR #1351, opened 2025-09-16
leverage weightedstake in aggregation

libotony:
# Description

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing E2E tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have not added any vulnerable dependencies to my code

PR #1386, opened 2025-09-24
fix(ci): build solo first then run

darrenvechain:
# Checklist:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -82,8 +82,11 @@ jobs:
         with:
           go-version: 1.24.x
 
+      - name: Make all
+        run: make all
+
       - name: Start Solo
-        run: nohup go run ./cmd/thor ${{matrix.args}} > thor-solo.log 2>&1 &
+        run: bin/thor ${{matrix.args}} > thor-solo.log 2>&1 &
 
       - name: Wait for block 1
         run: |


libotony:
I remember this was introduced in a PR targeting master and then got reverted?

darrenkel:
> I remember this was introduced in a PR targeting master and then got reverted?

`release/hayabusa` but I don't see it there https://github.com/vechain/thor/blob/release/hayabusa/.github/workflows/test.yaml

libotony:
Here https://github.com/vechain/thor/pull/1353 merged into PR https://github.com/vechain/thor/pull/1351, but got reverted in this [commit](https://github.com/vechain/thor/pull/1351/commits/1698ed2361a8cc39f443c2fff672e2769aaefd06).

otherview:
I believe I reverted this. I think this was created because the CI was flaky ? But I don't see how this can improve it.

darrenkel:
> I believe I reverted this. I think this was created because the CI was flaky ? But I don't see how this can improve it.

Its flaky because we time out after 30 seconds. In that 30 seconds we're currently building and then starting. If the build takes more than 30 seconds, test fails. This change ensures the build is not part of the 30 seconds.

otherview:
> > I believe I reverted this. I think this was created because the CI was flaky ? But I don't see how this can improve it.
> 
> Its flaky because we time out after 30 seconds. In that 30 seconds we're currently building and then starting. If the build takes more than 30 seconds, test fails. This change ensures the build is not part of the 30 seconds.

That's a good point. So we don't care about the build time, just the run time for timeouts.

.github/workflows/test.yaml
```diff
@@ -82,8 +82,11 @@ jobs:
         with:
           go-version: 1.24.x
 
+      - name: Make all
+        run: make all
```
otherview: Already covered above.

```suggestion
        run: make thor
```

PR #1390, opened 2025-09-29
Vanja/feat/836 ingest without index and journal

vanja-vechain:
# Description

Sync without indexes and with journal off

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- [x] Test A
- [x] Test B

**Test Configuration**:

* Go Version:
* Hardware:
* Docker Version:

# Checklist:


cmd/thor/node/packer_loop.go
```diff
@@ -211,3 +219,24 @@ func (n *Node) pack(flow *packer.Flow) (err error) {
...
+	if err != nil {
+		logger.Error("Error while switching journal mode", "err", err)
+		return err
+	}
+	err = writer.CreateIndexes()
+	if err != nil {
+		logger.Error("Error while creating indexes", "err", err)
+		return err
+	}
+	if err != nil {
```
paologalligit: I think this is a duplicate from above

cmd/thor/node/node.go
```diff
@@ -292,6 +297,12 @@ func (n *Node) guardBlockProcessing(blockNum uint32, process func(conflicts uint
 		conflicts []thor.Bytes32
 	)
 
+	if n.completingSync {
```
paologalligit: small thing: what if we had something like

```
func (n *Node) guardBlockProcessing(blockNum uint32, process func(conflicts uint32) (thor.Bytes32, error)) error {
    if n.completingSync {
        <-n.syncCompleteCh
    }

	n.processLock.Lock()
	defer n.processLock.Unlock()

	var (
		err       error
		blockID   thor.Bytes32
		conflicts []thor.Bytes32
	)
```
so that we don't lock - unlock - and relock?

logdb/logdb.go
```diff
@@ -59,7 +59,15 @@ func New(path string) (logDB *LogDB, err error) {
...
 	}
 
-	if _, err := wconn2.ExecContext(context.Background(), "pragma synchronous=off"); err != nil {
+	if _, err := wconn1.ExecContext(context.Background(), "PRAGMA journal_mode=WAL"); err != nil {
+		return nil, err
+	}
+
+	if _, err := wconn2.ExecContext(context.Background(), "PRAGMA journal_mode=MEMORY"); err != nil {
+		return nil, err
+	}
```
freemanzMrojo: I might be wrong on this, but If I remember properly even though we create 2 connections the `journal_mode` config applies to the DB (meaning that `MEMORY` would override `WAL`).

This is the source I found right now, I know, it is not crystal clear :( https://sqlite.org/pragma.html#pragma_journal_mode

Would it make sense to have a test for this?

cmd/thor/node/packer_loop.go
```diff
@@ -211,3 +219,24 @@ func (n *Node) pack(flow *packer.Flow) (err error) {
 		return newBlock.Header().ID(), nil
 	})
 }
+
+func (n *Node) completeSync() error {
+	writer := n.logDB.NewWriter()
+	logger.Info("Sync done, creating indexes")
+	err := writer.ExecuteJournalWalModeSwitch()
```
freemanzMrojo, line 198: Ok, maybe that is why this line exists? If that is the case, maybe it makes sense to just remove the WAL assignment at `New`.

cmd/thor/node/node.go
```diff
@@ -68,12 +68,14 @@ type Node struct {
...
-	logDBFailed   bool
-	initialSynced bool // true if the initial synchronization process is done
-	bandwidth     bandwidth.Bandwidth
-	maxBlockNum   uint32
-	processLock   sync.Mutex
-	logWorker     *worker
+	logDBFailed    bool
+	initialSynced  bool // true if the initial synchronization process is done
+	completingSync bool
+	syncCompleteCh chan struct{}
```
freemanzMrojo: Minor: Maybe it makes sense to encapsulate these 3 in a struct for `sync` (`SyncConfig`?) so we just add the minimum number of args.

freemanzMrojo (approved):
LGTM, left some comments

libotony:
@vanja-vechain Please also consider take this into account https://github.com/vechain/thor/blob/release/v2.4.1/cmd/thor/node/block_exec.go#L285-L292

logdb/logdb.go
vanja-vechain: Yes, you are correct and that is the behaviour that we are looking for, journal should be switched to WAL after the initial sync is done

cmd/thor/node/node.go
vanja-vechain: Makes sense, fixed

cmd/thor/node/packer_loop.go
vanja-vechain: Correct, it is slightly refactored now, please check

cmd/thor/node/packer_loop.go
vanja-vechain: Yes, fixed now

cmd/thor/node/node.go
vanja-vechain: Makes sense, applied

vanja-vechain:
> @vanja-vechain Please also consider take this into account https://github.com/vechain/thor/blob/release/v2.4.1/cmd/thor/node/block_exec.go#L285-L292

Yes, this is taken into account, without wal is actually executing for w = n.logDB.NewWriterSyncOff()

vanja-vechain:
Replace with https://github.com/vechain/thor/pull/1478

PR #1408, opened 2025-10-08
chore(energy): cap validator rewards to 100%

darrenkel:
# Description

cap validator rewards to 100%

# Checklist:

### DIFF
--- a/builtin/staker_native.go
+++ b/builtin/staker_native.go
@@ -186,8 +186,6 @@ func init() {
 			}
 
 			if !isPoSActive {
-				charger.Charge(thor.SloadGas) // a.getEntry(ValidatorMaster)
-
 				exists, endorser, _, _, err := Authority.Native(env.State()).Get(thor.Address(args.Validator))
 				if err != nil {
 					return nil, err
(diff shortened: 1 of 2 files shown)


builtin/energy/energy.go
```diff
@@ -285,7 +285,7 @@ func (e *Energy) DistributeRewards(beneficiary, signer thor.Address, staker stak
 
 	// If delegated amount of VET is 0 then transfer the whole reward to the validator
 	proposerReward := new(big.Int).Set(reward)
-	if hasDelegations {
+	if hasDelegations && validatorRewardPerc.Cmp(big.NewInt(100)) < 0 {
```
paologalligit, line 288: Does this mean that if a validator has delegators but the `validatorRewardPerc` for some reason is greater or equal 100, delegators are not getting any rewards?

darrenkel: Yes. Previously if amount was greater than 100, stargate would get negative rewards, which would cause the protocol to break I think (if stargate rewards was negative it can't be RLP encoded)

PR #1617, opened 2026-05-11
  Refactor eth tx nonce subsystem and enforce linear growth                                                                                                                                                           

libotony:
## Summary                                                                                                                                                                                                          
                                        
  Builds on `eth-equivalence` (which already has on-state nonce wiring from #1616 with two changes:   
                                                                                                                                                                                                                      
  **Refactor**                          
  - `state.Account.Nonce`: `[]uint64` with `rlp:"tail"` → `uint64` with `rlp:"optional"` (same byte-encoding, simpler in-memory shape).                                                                                                                                                   
  - `runtime/statedb`: struct-with-txType-field → `StateDB` interface with V1 (no-op SetNonce) and V2 (write-through SetNonce). Dispatched by `txCtx.Type` in `Runtime.newStateDB`.                                                                                                                                                                             
  - Nonce CALL-path inc moved from `PrepareTransaction` (post-clause) to `PrepareClause` (pre-`evm.Call`). Mirrors Ethereum's nonce-write-outside-EVM-snapshot pattern; on VMErr, eth tx skips the outer `state.RevertTo` so the increment survives.                                                                                                                                                                                        
                                                                                                                                                                                                                      
  **Add**                                                                                                                                                                                                             
  - Linear nonce growth enforcement across txpool / packer / consensus. `<` reject, `>` queue/defer/invalid, `==` admit.                                                                                                                                                                  
  - End-to-end eth tx execution tests (transfer / creation / revert preserves nonce / sponsored / fee floor / insufficient balance).                                                                                                                                                            
  - Multi-clause revert regression test.                                                                                                                                                                              
  - Cross-coverage with evm-upgrades EIPs (7825 gas cap, 1153 transient storage isolation, 6780 same-tx + pre-existing selfdestruct)

# Checklist:


moglu2017 (approved):

consensus/validator.go
```diff
@@ -299,7 +299,22 @@ func (c *Consensus) verifyBlock(blk *block.Block, state *state.State, blockConfl
 			}
 		}
 
-		receipt, err := rt.ExecuteTransaction(tx)
+		// Eth tx requires linear nonce growth (mismatch is a hard consensus error).
```
otherview, line 302: Doesn't this mean we'll need to run the nonce validation on every component ?

Wouldn't we want the nonce verification to happen at a core Method like `PrepareTransaction` or `ResolveTransaction` ?

vm/evm.go
```diff
@@ -458,6 +458,9 @@ func (evm *EVM) create(caller ContractRef, code []byte, gas uint64, value *big.I
 		return nil, common.Address{}, gas, ErrInsufficientBalance
 	}
 	nonce := evm.StateDB.GetNonce(caller.Address())
+	if nonce+1 < nonce {
```
otherview, line 461: Would that work too ?
```suggestion
	if nonce == math.MaxUint64  {
```

runtime/runtime.go
```diff
@@ -419,6 +421,14 @@ func (rt *Runtime) PrepareClause(
 			data, caddr, leftOverGas, vmErr = evm.Create(vm.AccountRef(txCtx.Origin), clause.Data(), gas, clause.Value())
 			contractAddr = (*thor.Address)(&caddr)
 		} else {
+			if txCtx.Type == tx.TypeEthDynamicFee {
+				nonce := stateDB.GetNonce(common.Address(txCtx.Origin))
+				if nonce+1 < nonce {
+					return nil, false, errors.New("nonce has max value")
```
otherview, line 419: Should we use the same error ?

```suggestion
					return nil, false, vm.ErrNonceUintOverflow
```

runtime/statedb/statedb.go
```diff
@@ -138,32 +187,15 @@ func (s *StateDB) AddBalance(addr common.Address, amount *big.Int) {
...
-func (s *StateDB) GetNonce(addr common.Address) uint64 {
-	if s.txType != tx.TypeEthDynamicFee {
-		return 0
-	}
-	n, err := s.state.GetNonce(thor.Address(addr))
-	if err != nil {
-		panic(err)
-	}
-	return n
-}
```
otherview, line 152: Catalogging this for a chat later - 

I like the approach, it pushes the logic upstream of wether the account should or should not have a Nonce.

This allows for a performance enhancement, no state -> cache pulling for accounts that can't possibly have a nonce.
(unless there's a hash colision, but even if that's the case, the nonce should not be updated because it would be going via the VechainTx path)

My concerns is that it's starting to look like all of this for one if statement.
 
What if these there was one statedb.New(), but with an optional param? 

Something like:
`New(state *state.State, nonceEnabled bool)` ?

paologalligit (approved):

otherview (approved):

PR #1619, opened 2026-05-25
Paolo/txpool metrics

paologalligit:
## Description

This updates txpool metrics so `txpool_current_tx_count` represents the current pool contents by `source`, `type`, and `status`, instead of mixing add/remove deltas with synthetic sources like `washed` or `n/a`. 

Txs now track their source explicitly — `local`, `remote`, or `fill` — and their status (`executable` / `non_executable`) is reflected on the gauge label. 

Wash removals decrement the current-count gauge using the removed tx's real labels and also increment a separate `txpool_washed_tx_count{source,type}` counter. 

`TxObject.executable` is now `atomic.Bool` so the metric helper can read it safely from concurrent `Remove` and `wash` paths. 

The internal `wash()` return surface is collapsed to a single `removed int`; the per-type breakdown now lives only on the counter. 

## NOTE — Backward-incompatible changes

Dashboards / alerts:

- `txpool_current_tx_count` gains a new `status` label (`executable` | `non_executable`). Existing grafana queries that sums or groups the gauge must use `sum without(status)` to reproduce the old totals; otherwise series will double-count or appear as separate rows.
- `source="washed"` and `source="n/a"` are no longer emitted on `txpool_current_tx_count`. Any panel or alert filtering on those values will go silent. The per-source/per-type wash rate now lives on the new `txpool_washed_tx_count{source,type}` counter — switch wash dashboards to `rate(txpool_washed_tx_count[...])`.
- Filled (sync-replayed) txs now report `source="fill"` instead of `source="remote"`.

# Checklist:

### DIFF
--- a/txpool/tx_object.go
+++ b/txpool/tx_object.go
@@ -8,7 +8,6 @@ package txpool
 import (
 	"math/big"
 	"slices"
-	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
@@ -43,9 +42,7 @@ type TxObject struct {
 	// gets <reward-ratio>% of the tip, after GALACTICA it's the effective priority fee per gas and validator gets 100% of the tip
 	priorityGasPrice *big.Int
 
-	// executable is updated by the pool. Accessed concurrently from wash, Remove and
-	// metric helpers
-	executable atomic.Bool
+	executable bool // don't touch this value, will be updated by the pool
 }
 
 func ResolveTx(tx *tx.Transaction, localSubmitted bool) (*TxObject, error) {
@@ -147,7 +144,7 @@ func (o *TxObject) Executable(chain *chain.Chain, state *state.State, headBlock 
 	}
 
 	// non executable -> executable, update payer, cost and priority gas price
-	if !o.executable.Load() {
+	if !o.executable {
 		o.payer = &payer
 		o.cost = prepaid
 
--- a/txpool/tx_object.go
+++ b/txpool/tx_object.go
@@ -8,6 +8,7 @@ package txpool
 import (
 	"math/big"
 	"slices"
+	"sync/atomic"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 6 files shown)


txpool/tx_object.go
```diff
@@ -20,23 +21,42 @@ import (
...
 	cost           *big.Int      // total tx cost the payer needs to pay before execution(gas price * gas)
 
 	// basic unit of tip price for the validator, before GALACTICA it's the overallGasPrice(provedWork included) and validator
 	// gets <reward-ratio>% of the tip, after GALACTICA it's the effective priority fee per gas and validator gets 100% of the tip
 	priorityGasPrice *big.Int
 
-	executable bool // don't touch this value, will be updated by the pool
+	// executable is updated by the pool. Accessed concurrently from wash, Remove and
+	// metric helpers
+	executable atomic.Bool
```
libotony: My understading is executable does not have a race condition as it is well managed in the pool, it's fine for the metrics handler to have a snapshot of this value, no need to mark it atomic.

paologalligit: There's a tiny chance of a race condition in:

- `wash()`: while running `Executable()` it writes `o.payer = &payer` and `o.cost = prepaid`.
- if `packer` packed that tx, then it calls `Remove()`: it calls `RemoveByHash`, which reads `txObj.Payer()` and `txObj.Cost()` under the map lock, but the `TxObject` fields aren't covered by that lock.

It looks a very rare case, but probably it should be addressed. I can revert the atomic bool and introduce this fix in a follow-up PR, if you prefere

txpool/tx_object.go
```diff
@@ -37,6 +46,14 @@ type TxObject struct {
...
 
 func ResolveTx(tx *tx.Transaction, localSubmitted bool) (*TxObject, error) {
+	source := txSourceRemote
+	if localSubmitted {
+		source = txSourceLocal
+	}
+	return resolveTxWithSource(tx, localSubmitted, source)
+}
+
+func resolveTxWithSource(tx *tx.Transaction, localSubmitted bool, source txSource) (*TxObject, error) {
```
otherview: Looks like the `localSubmitted` might be redudant here right ?
 
```suggestion
func resolveTxWithSource(tx *tx.Transaction, source txSource) (*TxObject, error) {
```

otherview (approved):
Minor bit with the duplication of field given that they are derived, lgtm otherwise

libotony (approved):

txpool/tx_object.go
paologalligit: I don't get why, can you expand a bit more?

libotony: Maybe Pedro means `localSubmitted = source=="local"`

paologalligit: ahhhh, right, that makes sense

libotony (approved):

PR #1644, opened 2026-07-24
Tidy up workflows and tests

otherview:
# Description

Small tidy ups found in older PR's

# Checklist:

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,10 +11,14 @@ permissions:
   contents: read
 
 jobs:
+  # Cross-platform build and test coverage. Linux is covered by the
+  # test_race and test_coverage jobs below.
   unit_tests:
+    name: unit_tests (${{ matrix.os }})
     strategy:
+      fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
@@ -30,7 +34,23 @@ jobs:
 
       - name: Make Test
         id: unit-test
-        run: make clean test
+        run: make test
+
+  # Race detection runs on arm64: linux/arm64 is a release target, and its
+  # weaker memory model can surface races that amd64 hides.
+  test_race:
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Install Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Make Test Race
+        run: make test-race
 
   test_coverage:
     runs-on: ubuntu-latest
@@ -42,9 +62,6 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
-
-      - name: Make all
-        run: make all
 
       - name: Make Coverage
         run: make test-coverage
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -31,7 +31,7 @@ jobs:
 
       - name: Make Test
         id: unit-test
-        run: make test
```
libotony, line 33: should that be `make clean all` first then `make test`?

otherview: Yeah that's a good idea. 
Now that I'm looking more into the workflow this clean at this stage is at a terrible place. Will look further into this flow - thanks !

otherview:
### Update: test execution tidy-up

- Dropped the `make clean test` approach - determinism now comes from `-count=1` in the
  `test` / `test-coverage` targets.
- Race detection split into its own `test_race` job + `make test-race`; `-race` removed from
  `test-coverage`. 
  - A race failure and a Codecov failure are no longer the same red X.
  
- `test_race` runs on `ubuntu-24.04-arm`: linux/arm64 is a release target and its weaker
  memory model can surface races amd64 hides. This is the first arm64 race run, so a failure
  here is more likely a real find than a flake.
  
- `ubuntu-latest` dropped from the `unit_tests` matrix - it duplicated `test_coverage`
  (same OS/arch). 
  
- `make all` removed from `test_coverage`; Linux linking is already covered by the Dockerfile
  build and `solo-test`.
  
- `go tool cover -html` moved to a new `test-coverage-html` target - it was running on the
  runner with no browser. Locally, use `make test-coverage-html`.
  
- Matrix job name pinned to `unit_tests (${{ matrix.os }})` so the check name survives Go
  bumps (it changed with every one: 1.23 → 1.24 → 1.25.4 → 1.25.x → 1.26.x), plus
  `fail-fast: false` so macOS and Windows both report.

Makefile
```diff
@@ -63,7 +63,10 @@ clean: #@ Clean the test and build cache and remove binaries
 	@echo "done. build cache and binaries removed."
 
 test:| go_version_check #@ Run the tests
-	@go test -cover $(PACKAGES)
+	@go test -count=1 -cover $(PACKAGES)
```
libotony: This will cause local test runs all tests all the time, it might be prefferred locally to have test results cached for untouched packages. Is there another way for having clean tests in CI?

.github/workflows/test.yaml
```diff
@@ -11,11 +11,15 @@ permissions:
...
 jobs:
+  # Cross-platform build and test coverage. Linux is covered by the
+  # test_race and test_coverage jobs below.
   unit_tests:
+    name: unit_tests (${{ matrix.os }})
     strategy:
+      fail-fast: false
       matrix:
         go-version: [1.26.x]
-        os: [ubuntu-latest, macos-latest, windows-latest]
```
cristidas, line 17: Why was `ubuntu-latest` removed from here?

Makefile
otherview: Yes, I considered that would be ok, the unit tests are pretty fast.
Would could leave it there or I could have the CI run the `-count=1` explicitly. Do you have a preference ?

.github/workflows/test.yaml
otherview: It was a duplicated - the coverage test already runs the tests using ubuntu-latest.

libotony (approved):

libotony (approved):

paologalligit (approved):

PR #1618, opened 2026-05-15
Adds EIP-2935 HISTORY_STORAGE contract support

moglu2017:
# Description

Introduces a historical block hash contract with a fixed address per EIP-2935, enabling dApps using the EIP-2935 convention to function seamlessly. Loads the contract code at the fork activation block to ensure compatibility and interoperability.

The different with Ethereum
1. Thor's contract only implements the `Get` logic, redirect to the executor builtin contract.
2. The contract's codehash is different.
3. The contract's nonce is 0, etherem is 1.

Fixes # (issue)

## Type of change
- [x] New feature (non-breaking change which adds functionality)

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] New and existing E2E tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules
- [x] I have not added any vulnerable dependencies to my code

PR #1645, opened 2026-07-28
Ensure Gas Pricing + Fix Unused Helper + Tweak Comments

otherview:
# Description

Follow-ups to #1618 (EIP-2935 HISTORY_STORAGE). Tests and documentation only -
no change in contract behaviour.

- **Make reverting-clause gas observable** (`test/testchain/clause_call.go`)
`ClauseCall` returned `nil, 0, err` on a VM revert, discarding the gas the
clause had burned. Extracted `ExecClause`, which returns
`ClauseResult{Data, GasUsed, VMErr}` and keeps `GasUsed` valid on revert.

- `ClauseCall` is now a thin wrapper preserving its exact previous contract, so
all existing callers are untouched.

- **Pin the facade's gas on every path** (`builtin/history_test.go`)
Collapsed three near-duplicate call helpers into one — `callHistoryRawWithGas`
returned a gas value that every caller threw away — and asserted it:

| Input | Gas | Guard |
|---|---|---|
| value > 0 | 45 | callvalue (runs first) |
| calldatasize != 32 | 78 | length |
| num >= block.number | 215 | future |
| block.number - num > 8191 | 297 | serve window |
| valid in-window read | 2085 | success |

- Clause-execution gas only, excluding intrinsic transaction gas.
invariant across fork config, chain depth, target block number,
at the activation block itself. The figures are distinct per guard, so a
failure identifies which check fired rather than just "it revert
catch the fallback accidentally becoming `payable`.

- **Document the non-payable fallback as deliberate** (`history.sol`, `builtin.go`)
The old comment claimed dApps "observe identical failure semantics on Thor".
That isn't true for value-bearing calls: the EIP-2935 reference bytecode has no
CALLVALUE guard, so mainnet accepts the funds and returns the hash where we
revert. Keeping non-payable — it matches every other Thor builtin, and rejects
a caller bug instead of silently absorbing funds. `staticcall` reads cannot
carry value and are unaffected.

# Checklist:


runtime/runtime.go
```diff
@@ -145,6 +145,16 @@ func New(
 		}
 	}
 
+	// EIP-2935 HISTORY_STORAGE facade. The contract reads historical block IDs
+	// straight from the Extension builtin at call time, so deploying the code is
+	// all that is needed — unlike the mainnet reference contract there is no
```
libotony:
```suggestion
	// all that is needed — unlike the ethereum reference contract there is no
```

runtime/runtime.go
```diff
@@ -145,6 +145,16 @@ func New(
...
+	// storage ring buffer to seed here, and none to maintain per block.
+	if forkConfig.INTERSTELLAR == ctx.Number {
```
libotony: Any considerations to move this code here?

otherview: Shouldnt have been moved. The branching was all wonky.

moglu2017 (approved):

paologalligit (approved):

libotony (approved):

PR #1677, opened 2026-09-22
Check the release signing subkey expiry daily

libotony:
# Description

Adds `check-signing-key-expiry.yaml`, a daily workflow that reads the signing subkey expiry from `docs/release-signing-key.asc` and notifies Slack at 15, 7, 3 and 1 days remaining, then every day once expired. The run summary shows the subkey, its expiry date and days left.

The check reads the committed public key, so a subkey rotation must also commit the refreshed `.asc`.

## Type of change

# How Has This Been Tested?

- Ran the parsing step locally against the committed key: expires 2027-09-02, notify=false.
- Verified the empty-input path reports `::error` and fails.
- `actionlint` passes.


otherview (commented):
`case "$days" in 15|7|3|1)` only fires on exact day matches. Scheduled workflows on GitHub Actions are best-effort and can occasionally get delayed or skipped - if the day-15 run doesn't fire, that milestone is gone for good, and we won't hear about it again until day 7.

Suggest switching to a threshold instead of exact matches, e.g. weekly starting at 30 days, then daily inside the last week:
```
if [ "$days" -le 7 ] || { [ "$days" -le 30 ] && [ $(( days % 7 )) -eq 0 ]; }; then
  notify=true
fi
```

That way a missed run just gets caught on the next day's check instead of silently skipping a milestone.

.github/workflows/check-signing-key-expiry.yaml
```diff
@@ -0,0 +1,91 @@
...
+            echo "## Release signing key"
+            echo
+            echo "| Signing subkey | Expires on | Days left |"
+            echo "|---|---|---|"
+            echo "| \`$keyid\` | **$expires_on** | $days |"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+          if [ "$days" -le 0 ]; then
+            echo "::error::signing subkey expired on $expires_on"
+            exit 1
```
cristidas, line 67: @libotony do we want to also notify when this error happens?

libotony: It does. `notify=true` is set whenever `days <= 0`, and the `notify-slack` job runs with `if: always()`, so the Slack message still goes out when this step fails; the `exit 1` is only there to turn the run red in Actions as well.

The other error path, "no signing subkey with expiry found" (e.g. a malformed `.asc` was committed), did not notify. ba1c202 makes that one notify too, since a broken key file is exactly when we'd want to hear about it.

libotony:
@otherview Good point, an exact-day match does lose the milestone if that run is skipped. ba1c202 switches to your threshold form: daily inside the last 7 days, weekly between 8 and 30 days. The weekly tier can still miss a single day (28/21/14), but the worst case is then a week's delay followed by daily alerts, which is fine for a key that we rotate months in advance.

otherview (approved):

cristidas (approved):

PR #1678, opened 2026-09-22
fix: txpool metric double count on race

otherview:
# Description

`thor_metrics_txpool_current_tx_count` can drift permanently upward from the real pool size under normal, expected conditions (multiple peers gossiping the same transaction at nearly the same time) — with no corresponding growth in the actual `TxPool`. On a busy mainnet node we found the gauge reporting ~372,000 "non-executable" transactions while `/node/txpool/status` showed the real pool held 5.

#### Root cause

In `txpool/tx_pool.go`, `add()` uses an unlocked pre-check to skip work for true duplicate resubmissions:

```go
if p.all.ContainsHash(newTx.Hash()) {
    return nil
}
``` 
If two goroutines process the same tx hash concurrently (routine on a well-peered node), both can pass this check before either has inserted. The mutex inside txObjectMap.Add() correctly serializes them and only inserts once — but Add() returned a bare error, nil both for "actually inserted" and for "hash already existed, no-op." Both callers therefore saw err == nil and both called addTxPoolMetric(txObj, 1). Only one real object ever exists to be removed later (one future -1), so every extra increment is a permanent, unrecoverable leak.

This is metrics-only: the actual pool (p.all), and therefore /node/txpool and /node/txpool/status, were never affected — verified live against a mainnet node.

### Fix

txObjectMap.Add() now returns (inserted bool, err error), threaded up through addWhenSynced/addWhenNotSynced. add() only fires the metric (and the addedAfterWash wash trigger) when inserted is true, making the metric's increment gated by the same lock that decides real insertion — closing the race.

Fixes # [(issue)](https://github.com/vechain/protocol-board-repo/issues/1234)

## Type of change

# Checklist:


txpool/tx_object_map.go
```diff
@@ -39,27 +39,32 @@ func (m *txObjectMap) ContainsHash(txHash thor.Bytes32) bool {
 	return found
 }
 
+// Add inserts txObj, returning whether it was newly inserted (false + nil
```
libotony: Can you make this comment generic other than talking one thing specificlly?

txpool/tx_pool.go
```diff
@@ -258,17 +258,23 @@ func (p *TxPool) add(newTx *tx.Transaction, rejectNonExecutable bool, localSubmi
...
-		err = p.addWhenNotSynced(newTx, txObj)
+		inserted, err = p.addWhenNotSynced(newTx, txObj)
 	}
 	if err != nil {
 		return err
 	}
 
-	atomic.AddUint32(&p.addedAfterWash, 1)
-	addTxPoolMetric(txObj, 1)
+	// inserted is false when a concurrent add of the same hash won the race in
```
libotony: Same comment here, comment can be clearer, I have this with claude before

txpool/tx_object_map.go
otherview: Agree, fixed 👍

txpool/tx_pool.go
otherview: Fixed 👍

libotony (approved):

